PR #1, opened 2015-01-15
GeometryKit to functions and moving it to BCFoundation

pieteromvlee:
Connect to BohemianCoding/Sketch#3824


samdeane:
Ah, it didn't show up because BCFoundation wasn't attached to Waffle. Fixed that now.

BCGeometry.m
```diff
@@ -0,0 +1,133 @@
...
+}
+
+
+CGPoint BCPointBetweenPoints(NSPoint p1, NSPoint p2) {
+  return CGPointMake((p1.x + p2.x) / 2, (p1.y + p2.y) / 2);
+}
+
+CGFloat BCFloatMakeNotInfOrNan(CGFloat value) {
+  if (isinf(value) || isnan(value))
+    return 0.1;
```
samdeane: might make sense to use CGFLOAT_MIN here

BCGeometry.m
```diff
@@ -0,0 +1,133 @@
...
+  else
+    return value;
+}
+
+CGFloat BCFloatValidValue(CGFloat value) {
+  value = BCFloatMakeNotInfOrNan(value);
```
samdeane: Not quite sure what the point of the MAX/MIN bits are here.

BCGeometry.m
```diff
@@ -0,0 +1,133 @@
...
+    for (NSInteger i = 1; i < 10; i += 2) {
+      if (slope > -M_PI * i / 8)
+        return -M_PI * (i - 1) / 8;
+    }
+  }
+  return slope;
+}
+
+
+CGPoint BCPointBetweenPoints(NSPoint p1, NSPoint p2) {
```
samdeane: Feels like Midpoint might be a better name for this?

Source/BCCornerGeometry.m
```diff
@@ -0,0 +1,129 @@
...
+BOOL BCCornerIsInMidHorizontal(BCCorner corner) {
+  return corner == BCCornerMidLeft || corner == BCCornerMidRight;
+}
+
+BOOL BCCornerIsInMidVertical(BCCorner corner) {
+  return corner == BCCornerMidTop || corner == BCCornerMidBottom;
+}
+
+BCCorner BCCornerFlipByAxis(BCCorner corner, BCAxis axis) {
+  if (axis == BCAxisHorizontal) {
```
samdeane: fwiw a lot of these methods could be done with simple lookup tables.

```
static BCCorner kFlippedHorizontal = { BCCornerTopRight, BCCornerTopRight, ... etc};
static BCCorner kFlippedVertical = { BCCornerBottomLeft, BCCornerBottomRight, ... etc};
return (axis == BCAxisHorizontal) ? kFlippedHorizontal[corner] : kFlippedVertical[corner];
```

that sort of thing...

Source/BCCornerGeometry.h
```diff
@@ -0,0 +1,52 @@
+//  Created by Pieter Omvlee on 19/01/2015.
+//  Copyright (c) 2015 Bohemian Coding. All rights reserved.
+
+#import <Foundation/Foundation.h>
+#import "BCGeometry.h"
+
+typedef NS_ENUM(NSInteger, BCCorner) {
+  BCCornerNone        = -1,
```
samdeane: Don't really need to explicitly give numbers here - the enums will just increase by 1.
Also, -1 is a weird choice for none (0 would make more sense)

Source/BCCornerGeometry.h
```diff
@@ -0,0 +1,52 @@
...
+  BCCornerTopRightMask    = 1 << BCCornerTopRight,
+  BCCornerBottomLeftMask  = 1 << BCCornerBottomLeft,
+  BCCornerBottomRightMask = 1 << BCCornerBottomRight,
+  BCCornerMidTopMask      = 1 << BCCornerMidTop,
+  BCCornerMidLeftMask     = 1 << BCCornerMidLeft,
+  BCCornerMidRightMask    = 1 << BCCornerMidRight,
+  BCCornerMidBottomMask   = 1 << BCCornerMidBottom
+};
+
+#define BCCornerCount 8
```
samdeane: The best way to get a count of enums is to put the count as the last item in the enum. 

Also #define is evil ;)

Source/BCFoundationShared.xcconfig
```diff
@@ -0,0 +1,6 @@
+#include "../Configs/BCStaticLibrary.xcconfig"
+
+PRODUCT_NAME=$(TARGET_NAME)
+DEFINES_MODULE = YES
+GCC_PREFIX_HEADER =
+DEAD_CODE_STRIPPING = NO
```
samdeane: Why is this off exactly?

samdeane: Do we still need this?

Source/BCPointGeometry.m
```diff
@@ -0,0 +1,138 @@
...
+BOOL BCPointsEqualWithMargin(CGPoint p1, CGPoint p2, CGFloat margin) {
+  return BCDistanceBetweenPoints(p1, p2) < margin;
+}
+
+
+CGPoint BCPointInvert(CGPoint point) {
+  return CGPointMake(-point.x,-point.y);
+}
+
+CGPoint BCPointSubtractPoint(CGPoint point1, CGPoint point2) {
```
samdeane: Not a massive deal, but I'd be tempted to change a few names:

The cocoa naming convention (NSOffsetRect etc) seems to be to put the type at the end... helpfully.

BCPointAddPoint -> BCAddPoint
BCSubtractPoint
BCPointWithOffset -> BCOffsetPoint
BCPointBetweenPointsAt -> BCInterpolatePoints

Source/BCPointGeometry.m
```diff
@@ -0,0 +1,138 @@
...
+  if (ABS(me.x - other.x) < margin)
+    me.x = other.x;
+  if (ABS(me.y - other.y) < margin)
+    me.y = other.y;
+  return me;
+}
+
+CGPoint BCPointSnapToRectWithMargin(CGPoint point, CGRect rect, CGFloat margin) {
+  CGPoint __block result = point;
+  BCCornerEnumerate(^(BCCorner corner) {
```
samdeane: If performance is a consideration here, it would make more sense to write this out longhand for each corner.
Also, do you really want to snap the mid point corners too?

Source/BCCornerGeometry.h
samdeane: Also, aren't these really handles and not corners?

Source/BCRectGeometry.h
```diff
@@ -0,0 +1,54 @@
+//  Created by Pieter Omvlee on 19/01/2015.
+//  Copyright (c) 2015 Bohemian Coding. All rights reserved.
+
+#import <Foundation/Foundation.h>
+#import "BCGeometry.h"
+#import "BCCornerGeometry.h"
+
+CGRect BCRectWithSizeProportionallyInRect(CGSize size, CGRect rect);
```
samdeane: Again, could possibly make some of these function names a closer match to the Cocoa ones. If we think it matters enough...

Source/BCRectGeometry.m
```diff
@@ -0,0 +1,352 @@
+//  Created by Pieter Omvlee on 19/01/2015.
+//  Copyright (c) 2015 Bohemian Coding. All rights reserved.
+
+#import "BCRectGeometry.h"
+#import "BCPointGeometry.h"
+
+CGRect BCRectWithSizeProportionallyInRect(CGSize size, CGRect rect) {
+  CGRect result;
```
samdeane: fwiw, you can calculate the aspect ratio once here, compare it > 1.0, then either multiply or divide by it.

the last two lines in the if/else clause are also identical...

samdeane:
Reviewed: added a few comments. I know that this code was old, but moving it is the right time to clean it up in my opinion (otherwise, when?).

There are quite a few inconsistencies, eg using NSWidth in some places and not others, using our own functions in some places not others, a few magic numbers in there, etc...

I think it could do with an actual spring clean - maybe make an issue for it if you're not wanting to do it now.

PR #8, opened 2015-05-27
Feature/4691

opsGavin:
Connect to BohemianCoding/Sketch#4691

Added a category to NSMapTable to do enumeration and subscripting in the same way as NSDictionary.


pieteromvlee:
Review: Very surprising that MapTable didn't have this stuff yet. Good addition

PR #13, opened 2015-07-01
Review 5245

pieteromvlee:
Fixes for 5245

Connect to BohemianCoding/Sketch#5245.


samdeane:
Reviewed: yep - makes a lot of sense to use FLT_EPSILON generally.

samdeane:
Actually, I'm not sure.

Isn't FLT_EPSILON the smallest difference that can be expressed as a float?

In which case, `fabs(a - b) < FLT_EPSILON` will only be true if `a == b` would also return true.

Except that you're using FLT_EPSILON with a double. There is probably a DBL_EPSILON, which is what you should really be using.

samdeane:
So to summarise: you're using FLT_EPSILON with CGFloat, which is actually a double, so you really should use DBL_EPSILON.

As a side effect of using the wrong constant, FLT_EPSILON will actually be larger than the smallest difference that can be expressed, so doing the < comparison might actually work where `a == b` would have failed, but really that's only a side effect of the mistake.

If you want to do a comparison with a larger margin of error, you should probably do `< (DBL_EPSION * 2.0)` or 3.0 or whatever margin you want.

pieteromvlee:
Ah, it's a double of course. I'll see if I can change that. Whichever the case though, it's better than those 0.0001 I used to use. I'll see if DBL_EPSILON exists and use that instead

samdeane:
> Whichever the case though, it's better than those 0.0001 I used to use. I'll see if DBL_EPSILON exists and use that instead

Sort of - but my point is that doing (a - b) < DBL_EPSILON is identical to doing a == b

pieteromvlee:
Ah, of course... :-/
So what would we compare two floats with to make sure they're (nearly) identical?

samdeane:
As I said above, I think if you want a small-but-not-0 constant to use, then it probably makes sense to make it a fix multiple of DBL_EPSILON. Whether it’s x2, x5, x10 or whatever just depends a bit on the level of accuracy/error we want to allow.

PR #14, opened 2015-07-02
Added new BCCache class based on GCD.

drewmccormack:
Connect to BohemianCoding/Sketch#5492


Tests/BCCacheTests.m
```diff
@@ -0,0 +1,91 @@
...
+- (void)tearDown {
+  [super tearDown];
+  cache = nil;
+}
+
+- (void)testAddingAndRetrievingACachedObject {
+  NSString *owner = @"Owner";
+  NSString *cachedObject = @"String";
+  [cache setCachedObject:cachedObject forOwner:owner];
+  ECTestAssertIsEqual([cache cachedObjectForOwner:owner], cachedObject);
```
pieteromvlee: I'm surprised this passes, as the setCachedObject is an async method. Could this cause random failures in the future perhaps on our test machine?

drewmccormack: Hmm, it's a good point. I suspect it passes because the write must be performed serially, and it is queued first. So it is dequeued first, and no read can start until it is finished. 

I think the behavior is guaranteed by dequeueing order and dispatch_barrier, but it does rely on an implementation detail, which is not nice. Maybe I could add a method to wait for the queue to flush, and just use that in testing. (Bloody tests always make you ruin your nice APIs :) )

samdeane: Surely the sync cachedObjectForOwner is guaranteed to run after the setCachedObject, so in that sense the test is fine.

drewmccormack: Yeah, I changed my mind on this as well. I think we can establish that this cache has the property that reads that come after a write will always get the newly written value, so it is more than an implementation detail. So test is fine, indeed.

samdeane:
Reviewed:

The class is nice, but I'm not sure about how useful it is, as it stands.

It's actually effectively doing some sort of lock/wait for every read operation, because we're using dispatch_sync, which has to wait for the queued operation to complete. 

In theory the blocks inside the dispatch_sync can run concurrently, but they are so simple that we've almost serialised access to the cache, and I wonder what the performance characteristics will be.

This approach would really come into its own if retrieving values from the cache was a more expensive operation. I think that's what Mike was referring to in his article when he said: 

> The gain for such a simple dictionary isn't too compelling, but for more complicated use cases, where the readers need to carry out an expensive sequence of atomic operations, it can make it easy to write fast, safe concurrent code.

The other issue I have with this design is that just having a weak key on the object isn't enough I think.

There are situations where we'll need more than one value cached for the same object. That's going to force us to store a dictionary in the cache, and then we'll need some sort of locking mechanism for accessing the values in it - so we're back to square one again.

If we're going to stick with this approach, I'd be tempted to extend the api to include a secondary key, so we effectively have:
- `getCachedValueWithKey:forObject:`
- `setCachedValue:withKey:forObject:`

and then we manage the nested dictionary. That way we can at least perform the dictionary lookup inside that sync block, and get the value of some parallelism whilst doing so.

The other thing we've lost from our current design is the ability to pass a block into the setter and have it invoked to calculate the value to store, if it's missing. The reason that this is important is that the work to create the value is the slow bit, and potentially every thread can hit the point where they need it calculated at the same time. If you do it by reading the cache, calculating the value, then writing to the cache, potentially every thread does the calculation. The first one to finish gets to set the cached value, but by that time the others are already calculating it too. If you pass in the worker block, you can ensure that it's only invoked once.

samdeane:
Also, with our old implementation we used a spin-lock per object, which is a bit expensive with memory, but minimises contention when accessing cached values for different objects. Here we're going to have a single one of some sort for the queue, which is contended over by every read on any object. 

I expect that adding things to the queue is an atomic compare-and-swap sort of thing which is virtually free, but waiting for a sync job to complete might be slower. It depends a bit I suppose if the implementation is smart enough to just run the block there and then, or if it actually has to put it on the queue and wait for a potentially different thread to execute it.

Source/BCCache.m
```diff
@@ -0,0 +1,105 @@
...
+  // Check for cached value
+  id object = [self cachedObjectForOwner:owner andKey:key];
+  if (object) {
+    return object;
+  }
+  
+  // Nothing in the cache, so create a new object with the block.
+  // Obtain a semaphore, then check the cache again, in case another thread was ahead of us and cached a value.
+  // If nothing is available, create a new object, and cache that.
+  NSUInteger semaphoreIndex = (unsigned long long)owner % BCCacheObjectCreationSemaphorePoolSize;
```
samdeane: One of the problems we had before, when I was incorrectly trying to pool the spin locks, was that the block which calculated the value to set can often also access the cache, leading to deadlock. Are we in danger of the same thing here?

drewmccormack: Yes, that is a good point. So I guess my options are @synchronized(self), assuming the same object will not try to cache in the block, or NSRecursiveLock. Correct?

samdeane:
Review 2: looks good. It's probably too risky to switch over to using this in 3.4, but let's merge the class into develop, and then we can experiment with it (maybe in the experimental tree?).

PR #16, opened 2015-07-14
Review 5632: Rect bounds can be 1px off if sublayers are not on small decimal numbers

pieteromvlee:
Fixes for 5632: Rect bounds can be 1px off if sublayers are not on small decimal numbers

Connect to BohemianCoding/Sketch#5632.


samdeane:
Reviewed: not sure that "step" is really the correct word to use here, but currently I can't think of a better one.

PR #18, opened 2015-08-27
Review 5806

pieteromvlee:
Fixes for 5806

Connect to BohemianCoding/Sketch#5806.


Source/BCRectGeometry.h
```diff
@@ -56,4 +56,5 @@ CGFloat BCRectValueForKey(CGRect rect, NSString *key);
 CGRect BCRectNormalise(CGRect rect);
 CGRect BCRectWithSizeOnAxisAspectRatio(CGRect rect, CGFloat size, BCAxis axis, CGFloat aspectRatio);
 CGRect BCRectMoveCornerToPoint(CGRect rect, BCCorner corner, CGPoint point);
-CGRect BCRectByEnsuringMidIsContainedInRect(NSRect rect, NSRect outerRect);
\ No newline at end of file
+CGRect BCRectByEnsuringMidIsContainedInRect(NSRect rect, NSRect outerRect);
+CGRect BCRectMoveInRect(CGRect innerRect, CGRect outerRect);
```
samdeane: A headerdoc comment would be useful here, as I'm having trouble understanding quite what it's supposed to do.

Also (if I've understood correctly), I think the name should be `BCRectMoveIntoRect`?

Tests/RectGeometryTests.m
```diff
@@ -51,4 +51,21 @@ - (void)testFloatByRounding {
   XCTAssertEqualWithAccuracy(0.01, BCFloatRoundToStep(number, 0.001), 0.001);
 }
 
+- (void)testMoveRectInRect {
+  NSRect innerRect = NSMakeRect(100, 100, 100, 100);
```
samdeane: Not quite understanding the naming of innerRect/outerRect - the outer one seems to be smaller than the inner one?

I think it may just be that you need to use different names for your parameters (and variables in the test) - eg `rectToMove` and `referenceRect`.

samdeane:
Reviewed: the code is probably fine, but the names are confusing.

pieteromvlee:
Fair point

Source/BCRectGeometry.h
pieteromvlee: You did understand correctly

Source/BCRectGeometry.h
```diff
@@ -56,4 +56,11 @@ CGFloat BCRectValueForKey(CGRect rect, NSString *key);
 CGRect BCRectNormalise(CGRect rect);
 CGRect BCRectWithSizeOnAxisAspectRatio(CGRect rect, CGFloat size, BCAxis axis, CGFloat aspectRatio);
 CGRect BCRectMoveCornerToPoint(CGRect rect, BCCorner corner, CGPoint point);
-CGRect BCRectByEnsuringMidIsContainedInRect(NSRect rect, NSRect outerRect);
\ No newline at end of file
+CGRect BCRectByEnsuringMidIsContainedInRect(NSRect rect, NSRect outerRect);
+
+/**
+ Returns a rect by basically moving innerRect into the outerRect if necessary.
```
samdeane: Still a bit confusing/ambiguous. 

if `innerRect` is bigger than `outerRect`, how can it be moved "into" it? 

Does this really mean that `innerRect` is moved so that it overlaps with `outerRect`, or so that as much of it as possible is contained by it?

I'm being pedantic, but it's genuinely not clear, and we're only going to get code reuse if people actually understand what the existing methods do :)

pieteromvlee: The code would not work if innerRect is bigger than outerRect. A situation that never could happen in the original code, where innerRect was the popover's window frame and outerRect was the screen's frame... :-/

samdeane: It's ok if the restriction is that innerRect has to be smaller than outerRect to work - all I'm saying is maybe the comment should state that (and maybe we should assert it)?

Otherwise there's a danger of doing that thing where a method is presented as if it's general purpose, but has been designed solely with reference to a single place that we know is going to call it. Which isn't a great way to build libraries...

pieteromvlee: I can completely see what you mean and I'm happy to either assert or make a note in the docs. My preference would be the latter I think maybe. Which one would you pick?

samdeane:
Given a choice I'd pick an assertion, since it does something actively if you break it. Your call though.

pieteromvlee:
Ah crap. ECCore isn't linked into BCFoundation. :(

PR #23, opened 2015-10-16
Review 6315

samdeane:
Fixes for 6315

Connect to BohemianCoding/Sketch#6315.


BCMachine.m
```diff
@@ -0,0 +1,65 @@
...
+inline NSData *BCGetMacAddress(void)
+{
+  mach_port_t masterPort;
+  kern_return_t result = IOMasterPort(MACH_PORT_NULL, &masterPort);
+  if (result != KERN_SUCCESS) {
+    ECLog( BCMachineChannel, @"Unable to get master port %@", @(result) );
+    return nil;
+  }
+
+  CFMutableDictionaryRef matchingDict = IOBSDNameMatching(masterPort, 0, "en0");
```
drewmccormack: I guess 'en0' exists on every Mac, right?

BCMachine.m
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  io_iterator_t iterator;
+  result = IOServiceGetMatchingServices(masterPort, matchingDict, &iterator);
+  if (result != KERN_SUCCESS) {
+    ECLog( BCMachineChannel, @"Failed create service iterator %@", @(result) );
+    return nil;
+  }
+
+  CFDataRef macAddressDataRef = nil;
```
drewmccormack: Very minor, but being a CF ref, NULL would be better than nil, I think.

BCMachine.m
```diff
@@ -0,0 +1,65 @@
...
+  io_object_t aService = IOIteratorNext(iterator);
+  if (aService == 0) {
```
drewmccormack: Better to use NULL than 0 in this code?

BCMachine.m
```diff
@@ -0,0 +1,65 @@
...
+      }
+      macAddressDataRef = (CFDataRef)IORegistryEntryCreateCFProperty(parentService, (CFStringRef)@"IOMACAddress", kCFAllocatorDefault, 0);
+      if (!macAddressDataRef) {
+        ECLog( BCMachineChannel, @"Failed to create IOMACAddress property" );
+      }
+      IOObjectRelease(parentService);
+    } else {
+      ECLog( BCMachineChannel, @"Failed get parent service %@", @(result) );
+    }
+    IOObjectRelease(aService);
```
drewmccormack: Never used this framework, so I'm not sure about this: Do we have to release aService? It is retrieved from an iterator, so there is no creation. Best check in any case.

BCMachine.m
```diff
@@ -0,0 +1,65 @@
...
+    aService = IOIteratorNext(iterator);
+  }
+  IOObjectRelease(iterator);
+
+  NSData *ret = nil;
```
drewmccormack: Is the purpose of this method to get the last entry? That seems to be what it currently does.
Is it enough to get the first entry, or any entry? Code could possibly be made a bit simpler in those cases.

BCTime.h
```diff
@@ -0,0 +1,6 @@
+//  Created by Sam Deane on 16/10/2015.
+//  Copyright (c) 2015 Bohemian Coding. All rights reserved.
+
+extern NSTimeInterval BCTimeRandomInterval(NSTimeInterval initialDelayInSeconds, NSUInteger rangeInSeconds);
```
drewmccormack: Find it a little bit odd that time is represented as both a double and an unsigned int in a single call. I think my preference would be to use NSTimeInterval for both arguments. I could imagine wanting to do sub-second ranges too.

drewmccormack:
Reviewed: A couple of questions.

BCMachine.m
samdeane: Yeah, that's from Apple's sample code. I was dubious myself originally, but it seems to work.

BCMachine.m
samdeane: Yeah, to be honest when I started using Objective-C many moons ago I made a decision that the distinction between nil and NULL was a bit pointless, and I preferred writing nil everywhere (I realise that technically they are distinct types).

It's not something we've generally bothered about in the Bohemian code, but you're right and we're trying to stay as Apple-like as we can, so I'll change it.

BCMachine.m
samdeane: Better to use `IO_OBJECT_NULL`!

BCMachine.m
samdeane: The Apple documentation is silent on this issue...

This is quite old code, and came original from an Apple sample - which doesn't mean it's correct of course, but I'm inclined to trust it.

BCTime.h
samdeane: Yes, I thought the same (I moved this code, then generalised it a bit so that it could be tested, but I didn't write the original).

It complicates the random number stuff a bit - I'm guessing that's why it was done that way - but it would be a more intuitive API. I'll change it.

PR #29, opened 2016-01-18
Review 7543: Implement non-UI springs and struts

opsGavin:
Fixes for 7543: Implement non-UI springs and struts

Connect to BohemianCoding/Sketch#7543.


Tests/RectGeometryTests.m
```diff
@@ -6,6 +6,8 @@
 
 #import "BCRectGeometry.h"
 
+static BOOL CGRectEqualWithApproximation( CGRect rect1, CGRect rect2 );
```
pieteromvlee: There is a `CHRectEqualToRect` function with a similar purpose. We can replace all uses of that with something like this. However, I would call it BCRectEqual, not CG.

pieteromvlee: Or move that down to foundation (which already has many rect functions) and use that everywhere

opsGavin: I've added `BCRectEqualToRectWithAccuracy` to Foundation and removed `CHRectEqualToRect` (which it turns out was never called from anywhere).

pieteromvlee:
Any reason you made it static inline whereas everything else in BCGeometry are normal functions?

opsGavin:
There were two reasons I left it static inline. Firstly this was more or less a copy of the old CHRectEqualToRect which I moved from Chocolat and that was static inline. But mainly I left it that way because that at the moment it's only accessed from the tests. I could make it a normal function, but then it would be built into the App even though it's not used.

Both fairly minor reasons though, so if you think it would be better as a normal function I'm happy to change it.

PR #35, opened 2016-02-15
Removed recursive lock from BCCache.

drewmccormack:
Connect to BohemianCoding/Sketch#7896


pieteromvlee:
Hmm nothing ever 'tried' the lock is it were so I can see why this could deadlock. :+1: we're on a serial queue anyway so makes sense to me

PR #42, opened 2016-03-15
Code Review: Cache operations can be duplicated on multiple threads (#8283)

drewmccormack:
Code review for Cache operations can be duplicated on multiple threads (#8283):

> I have a rectangle with a background blur which is 5000x4000. To render this layer we create a 5000x4000 bitmap, apply to blur and then render to a second 5000x4000 bitmap. This second image is then cached. The problem is that this work could be duplicated across many threads (e.g. in the example screenshot we are doing the same thing 6 times) and the performance hit is quite noticeable.
> 
> screen shot 2016-03-15 at 11 37 34
> 
> I understand from @drewmccormack that this is related to #7896 in that we used to have a lock around the operation, but that this was removed as it could cause a deadlock.

Connect to BohemianCoding/Sketch#8283.


opsGavin:
Reviewed: looks good

PR #43, opened 2016-03-30
Code Review: Rotation Stepper Can Change Signedness of Angle (#6389)

pieteromvlee:
Code review for Rotation Stepper Can Change Signedness of Angle (#6389):

> Steps to reproduce:
> 1. Make a new layer
> 2. Enter -45 in the transform box
> 3. Click the down stepper
> 4. Angle is now displayed as 314º
> 
> This behaviour is different to 3.3. It changed in #1819, where the rotation direction was reversed in the UI.

Connect to BohemianCoding/Sketch#6389.


drewmccormack:
Reviewed: I guess this function change won’t have side effects (?) Otherwise good.

pieteromvlee:
I checked, and its only used in those two places

PR #45, opened 2016-03-31
Code Review: Move foundation level code out of Chocolat into BCFoundation (#3763)

pieteromvlee:
Code review for Move foundation level code out of Chocolat into BCFoundation (#3763):

> Chocolat is a bit of a dumping ground, and it would be good to split it into smaller pieces.
> 
> There are various ways we could split it - either by functional area (eg date utilities, networking utilities, text utilities, etc...) or by the level of the Apple frameworks that it requires.
> 
> This issue is based on the second approach. Essentially we should move anything in Chocolat that only requires the Foundation classes into BCFoundation instead.

Connect to BohemianCoding/Sketch#3763.


drewmccormack:
Reviewed :+1:

PR #46, opened 2016-04-05
Code Review: Deadlock in BCCache (#8584)

opsGavin:
Code review for Deadlock in BCCache (#8584):

> #7739 has introduced a lock in the `BCCache`.
> 
> The problem is that the LL refresh code dispatches multiple blocks onto background queues which each want to access the cache. Within `BCCache` there is a `dispatch_sync` to the cache's queue which will block until a value has been read from the cache. When we need to set a value `dispatch_barrier_async` is called. Although this is asynchronous and the method will return, any subsequent `dispatch_sync` will block until the barrier has been reached. The problem is that the LL blocks will eventually all be waiting on the `BCCache`'s queue which will never be processed as GCD is using all it's threads to service the LL blocks.
> 
> This [WWDC video](https://developer.apple.com/videos/play/wwdc2015/718/) which @drewmccormack found explains the problem quite well. (Thread explosions - about 33mins in).
> 
> Although the LL refresh code is to blame for introducing this issue, it would be good to fix it in `BCCache` to prevent a similar issue in future. Discussing this with @drewmccormack, one solution is to replace the `dispatch_sync` / `dispatch_barrier_async` calls with a pthread `pthread_rwlock_t`.
> 
> /cc @pieteromvlee - I think this is probably a showstopper for 2016.4

Connect to BohemianCoding/Sketch#8584.


Source/BCCache.m
```diff
@@ -85,29 +85,29 @@ - (id)cachedObjectForOwner:(id)owner andKey:(id<NSCopying>)key orCreateWithBlock
...
+  [self->locker protectedWrite:^{
     [self->cachedObjectsByOwner removeAllObjects];
-  });
+  }];
 }
 
 - (NSUInteger)count {
   __block NSUInteger count = 0;
-  dispatch_barrier_sync(queue, ^{
+  [self->locker protectedRead:^{
```
opsGavin: I wasn't sure if this should be `protectedWrite`?. It was a `dispatch_barrier_sync` before, but it looks like it only needs read access?

drewmccormack: Read lock is probably OK here. As long as the container isn't being written, it should be safe, and so 'read' is enough.

drewmccormack:
Review: Code looks good. My only concerns are not really related to the new code per se:

1. This issue was triggered by exploding threads caused by LL use of GCD. A symptom was this problem in BCCache. But my fear is that the problem isn't really solved by this, just shifted. What will happen if some other part of the app tries to use GCD? Or even parts of a framework we have no control over? It seems to me we need the LL stuff under control as well.
2. I have no feel for whether this locking approach with pthreads is faster or slower than GCD. My guess is that it probably doesn't matter much in the grand scheme of things, but locks can be reasonably expensive, and GCD is designed to be at least partially lockless and more efficient. Do you have a feel for this?

opsGavin:
Good points.

Although the LL triggered the issue, I think the really the root of the problem was the combined use of `dispatch_sync` and `dispatch_barrier_async`. That said, it seems to be the standard approach to this problem (It seems to be the standard/only SO answer to this problem and it seems to be the one pushed by Apple) so it's entirely possible that we could run into it elsewhere.

I've pushed a change to the LL stuff which should help (in the main repo rather than here in BCFoundation). It's sort of a half-way house in that each layer will still push an individual block to GCD rather than a single block to update all layers at once. The reason I've left it like this is that the first implementation actually processed all layers in one block, but the code was really rather complex as we needed to go back to the main thread to update the preview images. This involved lots of maps/dictionaries to get back to the original layer from the immutable etc. It turned out there was no performance gain doing this and Pieter felt (and I agreed) that the simpler code of multiple dispatches would be better.

That said I have made a couple of changes which should still help with flooding GCD. The first is that I only dispatch if we have a shape. Previously any layer would trigger a dispatch even if it wouldn't do anything once it got there. The second is that I only dispatch if there is no cached bezier path on the immutable. I wasn't doing this before as accessing the cache from the main thread was causing problems, but it should be safe to do that now.

Regarding performance, I didn't notice any problems while testing, but I've added some performance tests to double-check. I was actually pleasantly surprised - it seems the pthread approach is between 4 to 8 times faster than GCD. I guess GCD can't avoid doing some sort of locking in this case and probably ends up falling back on pthreads.

drewmccormack:
Review 2: Oh, the performance tests are interesting. So we should definitely go with this solution. :+1: 

As an aside, something I have used in the past where a queue is overcome — e.g. the main queue — is an extra ‘throttling’ queue. For me it has always been a serial queue, but you could use NSOperationQueue with a maximum size.

The way it works is you submit to the throttling queue instead of the queue you are interested in. The throttling queue runs your block synchronously on the target queue. Effectively, it queues up tasks, and feeds them in to the other queue in a controlled way. I have found this works particularly well for the main queue, where you really don’t want excessive tasks arriving at once.

PR #47, opened 2016-04-22
Code Review: Ability to set Scale origin (#3991)

mikeabdullah:
Code review for Ability to set Scale origin (#3991):

> People would go nuts for this.
> 
> origin
> 
> How hard would it be to implement?

Connect to BohemianCoding/Sketch#3991.


Source/BCRectGeometry.h
```diff
@@ -63,6 +66,12 @@ NSComparisonResult BCRectCompare(CGRect rect1, CGRect rect2, BCAxis axis);
 CGFloat BCRectValueForKey(CGRect rect, NSString *key);
 CGRect BCRectNormalise(CGRect rect);
 CGRect BCRectWithSizeOnAxisAspectRatio(CGRect rect, CGFloat size, BCAxis axis, CGFloat aspectRatio);
+
+/**
+ Creates a rectangle the same size as \c rect, positioned so that \c corner is placed at \c point.
+ 
+ Creates a new rectangle rather than modifying an existing one as the name might suggest.
+ */
```
pieteromvlee: Are you suggesting we should rename this method? :-P

mikeabdullah: Yup! :-)

pieteromvlee: Then let's rename it :-P

mikeabdullah:
Review Notes: Spotted we had duplicate functions for making rectangles at a particular corner point. Furthermore, passing in an origin for an existing rectangle is pointless since we're about to rewrite it, so instead you now just pass in a size. Added test for the function.

pieteromvlee:
Review: 👍

PR #48, opened 2016-04-29
Code Review: Fix XC7.3 analyser warnings (#8772)

samdeane:
Code review for Fix XC7.3 analyser warnings (#8772):

> When attempting to use `boho stage` locally, the build fails after analysing with the following warnings
> 
> ````
>   - BCUndoManager.m, 55: Null is returned from a method that is expected to return a non-null value
>   - MSEventHandlerManager.m, 70: Null is returned from a method that is expected to return a non-null value
>   - MSEventHandler.m, 44: Null is returned from a method that is expected to return a non-null value
>   - MSAttributedString.m, 150: Null is returned from a method that is expected to return a non-null value
> ````
> 
> **Notes**
> 
> * No such warnings occur when building remotely on Jenkins
> * I've Xcode 7.3 (7D175) on my machine here
> * The above warnings occurred when attempting to stage a beta from `release/2016.3.1` branch

Connect to BohemianCoding/Sketch#8772.


mikeabdullah:
Please don't hate me too much, I corrected the header too, so the compiler actually sees it

samdeane:
Good call - but please, stop changing the code you're reviewing. I know you probably think that it's horribly inefficient not to just dive in and change things, but I think it's horrible rude *to* dive in and change things...

samdeane:
(it's trivial in this case I know, but I actually think that the principle is pretty important)

mikeabdullah:
Yes, I feel it's inefficient. What do you find rude about it? My logic is that the code is up for editing once it's merged in anyway, since we don't try to claim ownership of particular bits of the codebase.

samdeane:
> What do you find rude about it?

It feels a little arrogant and disrespectful to say "I am just going to change your code, because I know better than you". A review should be a conversation, and comments should be offered up in a spirit of humility. 

Yes the code is up for editing afterwards, but even then we have a collective process to decide what we're going to edit, and what we're going to do to it. 

In the context of a particular change that's already happened, I do think that it's quite important to say that the person who was given the responsibility for that piece of work has the ultimate say -- in all but exceptional circumstances.

The other I would say is that the sort of "efficiency" you're trying to improve here is arguably one that the world could do less of. There's a hint of "less haste, more speed". If the change is so trivial, it won't take the person who originally wrote the code long to do it either, and in the meantime they will have (a) found out about it and (b) had a chance to explain why it's not a good idea.

The final thing I'd say is that the whole point of the review process is that we're putting hats on, temporarily. The idea is that one person is writing the code, the other is checking / commenting / thinking-outside-the-box. That's the value of a review - a second pair of eyes on the changes. Once you start mixing the roles within a single piece of code, the value starts falling away. Am I supposed to review the change you made when reviewing my changes? How does that work?

As I said in my previous comment, I accept that this example is trivial, but I do think that the principle is quite important.

Going to cc/ @BohemianCoding/coders here, as I think this is something we need to agree on one way or the other, as a group.

pieteromvlee:
I agree with @samdeane here that keeping the roles of reviewer vs developer is better. I do remember past cases where what seemed like trivial changes turned out not to be - the reviewer had just misunderstood something smal but fundamental. Also, a reviewer's change might get missed by the developer and introduce an unwanted bug.

We are not peer-programming here, we're reviewing. Btw, this discussion has come up in #5350 as well and I thought the consensus there was also to keep the roles strictly separate. So let's do that. I appreciate the enthusiasm here but I think keeping proper separation is easier for all parties involved.

mikeabdullah:
> It feels a little arrogant and disrespectful to say "I am just going to change your code, because I know better than you". A review should be a conversation, and comments should be offered up in a spirit of humility.

My intent is more "hey, I think this would be even better, what do you think?"

> The other thing I would say is that the sort of "efficiency" you're trying to improve here is arguably one that the world could do less of. There's a hint of "less haste, more speed". If the change is so trivial, it won't take the person who originally wrote the code long to do it either, and in the meantime they will have (a) found out about it and (b) had a chance to explain why it's not a good idea.

Some of my feeling of inefficiency there is it may take a day or two for the author to be ready to come back round and deal with the review comments. At that point they might well respond to some comments with simply "I don't understand, can you explain more", and once this has gone all the way round, a week might have elapsed. Maybe I shouldn't worry about that.

> The final thing I'd say is that the whole point of the review process is that we're putting hats on, temporarily. The idea is that one person is writing the code, the other is checking / commenting / thinking-outside-the-box. That's the value of a review - a second pair of eyes on the changes. Once you start mixing the roles within a single piece of code, the value starts falling away. Am I supposed to review the change you made when reviewing my changes? How does that work?

I expect to say "ok, I've made my review and changes, now back to you to decide from there". I often find that when I come back and read my own code a little later, regardless of whether someone has edited it in the meantime or not, I do a more thorough job. And I think it's easy to disagree with a reviewer, and revert a commit, ideally with commentary as to why.

> We are not peer-programming here, we're reviewing. Btw, this discussion has come up in #5350 as well and I thought the consensus there was also to keep the roles strictly separate. So let's do that. I appreciate the enthusiasm here but I think keeping proper separation is easier for all parties involved.

I think this is the crux of my disagreement, and time for me to accept you don't want to work this way, and to let it lie :-)

PR #51, opened 2016-05-19
Code Review: Implement reading order (#9114)

opsGavin:
Code review for Implement reading order (#9114):

> There are various tickets we'd like to do that depend on being able to read the artboards in 'reading order'.
> 
> What we need is a fuzzy algorithm that returns an array of artboards that are in left-to-right-top-to-bottom order. Of course this isn't always possible, but assuming the user has done some logical ordering, we should be able to give a reasonable result. Of course if they're all in random locations then too bad.
> 
> Tickets that depend on this:
> 
> #6164 - Export Artboards to PDF in Reading Order
> #5543 - Easier Artboard Navigation with shortcuts in Reading Order
> 
> Perhaps we also want to show Mirror/Cloud artboards in Reading Order.
> 
> But first we need an algorithm :)

Connect to BohemianCoding/Sketch#9114.


pieteromvlee:
Cool stuff. And I appreciate the extensive comments and tests. 👍 
Let's hope QA likes it too :)

PR #55, opened 2016-06-06
Code Review: Implement an NSDictionary deserialiser (#9403)

opsGavin:
Code review for Implement an NSDictionary deserialiser (#9403):

> Write a deserialiser subclass for the deserialiser created by #9402
> 
> Part of #8880. The idea is that we'll have a serialiser which is producing JSON by this stage. We can use `NSJSONSerialization` to load this to a dictionary and then deserialise with this new class.
> 
> It may be that later we write our own JSON deserialiser class, but this class will provide a descent first step.

Connect to BohemianCoding/Sketch#9403.


Tests/BCJSONUnarchiverTests.m
```diff
@@ -0,0 +1,279 @@
...
+        same = [(BCJSONUnarchiverTestableClass*)self.object isNearlyEqual:other.object];
+      } else {
+        same = [self.object isEqualTo:other.object];
+      }
+    } else {
+      same = (other.object == nil);
+    }
+    return same && self.boolVal == other.boolVal &&
+      self.intVal == other.intVal &&
+      areAproxEqual( other.doubleVal, self.doubleVal ) &&
```
drewmccormack: Maybe ‘Approx’ for ‘Aprox'?

drewmccormack:
Review: Looks like good work. Small spelling comment, but otherwise good to merge into 8880.

PR #57, opened 2016-06-22
Code Review: Update our libraries to build as proper Clang modules (if possible) (#3845)

samdeane:
Code review for Update our libraries to build as proper Clang modules (if possible) (#3845):

> When modules first became available they were Apple only, but I think it’s now possible to define them for user code.
> 
> If true, it would be good to update our standard XCConfig files in BCFoundation so that all of our static libraries build as modules.
> 
> This ought to mean that we no longer have to add them to the link stage explicitly.

Connect to BohemianCoding/Sketch#3845.


BCFoundation.xcodeproj/xcshareddata/xcschemes/BCFoundation.xcscheme
```diff
@@ -3,16 +3,29 @@
    LastUpgradeVersion = "0730"
    version = "1.3">
    <BuildAction
-      parallelizeBuildables = "YES"
-      buildImplicitDependencies = "YES">
+      parallelizeBuildables = "NO"
+      buildImplicitDependencies = "NO">
```
mikeabdullah: Just checking, do we definitely need/want to turn both of these off?

Configs/BCShared.xcconfig
```diff
@@ -62,19 +62,23 @@ PRECOMPS_INCLUDE_HEADERS_FROM_BUILT_PRODUCTS_DIR = YES
...
+GCC_VERSION =
+GCC_C_LANGUAGE_STANDARD = compiler-default
+CLANG_CXX_LANGUAGE_STANDARD = compiler-default
+CLANG_CXX_LIBRARY = compiler-default
 
 GCC_ENABLE_CPP_EXCEPTIONS = YES
 GCC_ENABLE_CPP_RTTI = YES
-GCC_PREFIX_HEADER = Prefixes/BC$(TARGET_NAME)$(CONFIGURATION).pch
+GCC_PREFIX_HEADER = Source/$(TARGET_NAME).pch
+GCC_PRECOMPILE_PREFIX_HEADER = NO
```
mikeabdullah: So `GCC_PRECOMPILE_PREFIX_HEADER` is turned off for both debug and release now. What's the impact of this?

Tests/BCFoundationTests.pch
```diff
@@ -0,0 +1,38 @@
+//
+// Prefix header for all Mac-related source files
```
mikeabdullah: All Mac-related, or all test-related? Perhaps something else?

BCFoundation.xcodeproj/xcshareddata/xcschemes/BCFoundation.xcscheme
samdeane: Yes, I think so. It never gets the implicit dependencies right, and without that I can't see how it can parallelize the build. We could try just turning off the parallelize option, but on the whole I prefer that the workspace explicitly lists and builds any dependencies anyway, as that then serves as a form of documentation.

Configs/BCShared.xcconfig
samdeane: Subjectively, it seems not to have any impact. Now that we're building our libraries as modules, a pre-compiled binary representation of them gets loaded into the compiler at build time anyway - my understanding is that this is a cleaner version of pretty much the same thing as the precompiled headers.

Tests/BCFoundationTests.pch
samdeane: That comment is out of date. Will change.

samdeane: Having said that, we now only have Mac-related code using these headers, since Awkward have done their own thing :(

BCFoundation.xcodeproj/xcshareddata/xcschemes/BCFoundation.xcscheme
mikeabdullah: You can still have parallel builds with explicit dependencies I thought, no? Or is the problem that we don't have enough organisation made to set that up?

pieteromvlee:
👍

PR #58, opened 2016-06-23
Code Review: Float in Place doesn't work as expected if layer is hugging bound edges (#9673)

opsGavin:
Code review for Float in Place doesn't work as expected if layer is hugging bound edges (#9673):

> Consider this example: my rectangle layer and 'S' are grouped. The group is set to "Float in place". The 'Sketch' text layer is also set to Float.
> 
> Upon resizing the symbol, the 'float' group is actually pinned. I would have expected it to float in the middle, but as it bounds are on the left, top, and right edges of the symbol, it gets a little confused...
> 
> float
> 
> Would you agree that the expectation here would be for it to stick to the middle? If I wanted this result I would set it to Pin.

Connect to BohemianCoding/Sketch#9673.


pieteromvlee:
👍

PR #60, opened 2016-06-30
Code Review: Performance regression with smoothness of dragging text layers around canvas (#9670)

randomsequence:
Code review for Performance regression with smoothness of dragging text layers around canvas (#9670):

> Papadakis reported a noticeable lag when dragging text layers around in Sketch Beta 39 (31443).
> 
> I can confirm that this lag or apparent drop in frames is indeed particularly noticeable when comparing the current beta alongside Sketch 3.8.3.
> 
> **Steps**
> 
> 1. Insert a text layer into a new document
> 2. Select and drag the text layer around the canvas
> 
> **Notes**
> 
> * In a new document, 3.8.3 doesn't appear to drop any frames
> * In a direct comparison with 3.8.3, Sketch Beta 39 (31443) appears to drop some frames
> * Tested on a Retain display
> * Reproducible on a clean setup with no plugins installed
> 
> cc @pieteromvlee

Connect to BohemianCoding/Sketch#9670.


Source/BCObjectPool.h
```diff
@@ -0,0 +1,40 @@
...
+//  BCObjectPool.h
+//  BCFoundation
+//
+//  Created by Johnnie Walker on 30/06/2016.
+//  Copyright © 2016 Bohemian Coding. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+// BCObjectPool maintains a pool of arbitrary objects created with a common initialiser block.
```
samdeane: Trivial point - if you use `///` or `/**` here the comment should get aattached to the BCObjectPool class as documentation.

Source/BCObjectPool.m
```diff
@@ -0,0 +1,51 @@
...
+//  BCFoundation
+//
+//  Created by Johnnie Walker on 30/06/2016.
+//  Copyright © 2016 Bohemian Coding. All rights reserved.
+//
+
+#import "BCObjectPool.h"
+
+@interface BCObjectPool () {
+  NSHashTable *_pool;
```
samdeane: Any particular justification for using an NSHashTable here rather than, say, just an NSDictionary?

samdeane:
Reviewed: BCObjectPool looks good, a couple of trivial comments but nothing to hold it up.

Source/BCObjectPool.m
randomsequence: I don't want to use an `NSDictionary`, since I have no keys. I could use an `NSSet`, but that's problematic if the objects you're pooling return `YES` to `isEqual`, as the `NSTextStorage` objects I have do.

samdeane: Right.

samdeane: Probably worth adding a comment to that effect to the .m file, otherwise someone at a later date will change it and everything will break.

samdeane:
Ah, one thing I've found. You are #importing it incorrectly from within Sketch. 

You need to add `#import "BCObjectPool.h"` to the BCFoundation.h umbrella header. It will then get picked up for you as part of the module.

samdeane:
Same for `BCReadWriteLock.h` if it's not already in there.

Source/BCObjectPool.m
randomsequence: I've added a test to cover that - it'll break if someone were to replace it with `NSMutableArray`, for example.

Source/BCReadWriteLock.h
```diff
@@ -9,7 +9,7 @@
 #import <Foundation/Foundation.h>
 
 /// BCReadWriteLock is a wrapper arround pthread's pthread_rwlock_t
-/// providing concurrent read access with single thread read access.
+/// providing concurrent read access with single thread write access.
```
opsGavin: Ah, um, yeah, that wasn't right. Good catch ;-)

Tests/BCObjectPoolTests.m
```diff
@@ -0,0 +1,165 @@
...
+  ECTestAssertIntegerIsEqual(vendedObjects.count, count);
+  
+  [vendedObjects removeAllObjects];
+  
+  // now empty the pool again
+  for (NSUInteger i=0; i<count; i++) {
+    id object = [pool vendObject];
+    NSValue *objectValue = [NSValue valueWithNonretainedObject:object];
+    // verify this object was something we recycled
+    ECTestAssertTrue([recycledObjects containsObject:objectValue]); // HALP, this fails.
```
opsGavin: Not sure I understand the comment here - does it fail?

randomsequence: Oops, should've cleaned that up when it passed.

PR #62, opened 2016-07-04
Code Review: Ability to edit multiple subpaths in shape (#9169)

mikeabdullah:
Code review for Ability to edit multiple subpaths in shape (#9169):

> from #2577 
> 
> Creating tickets for the vector stuff we definitely want to work on. Idea here would be to press enter on a shape with sub paths, and you'll be able to enter all its children, without having to jump in and out of edit mode, selecting different layers.
> 
> A couple of caveats
> - Would this be complete chaos with a lot of sub-paths?
> - If a sub-path was selected, and sent to edit mode, would that layer be isolated, or would its siblings also be editable?
> 
> >Simple concept for multi-shape editing. All points, of all the selected shapes will be editable on edit, rather than just all the points of the topmost shape. This'll be nice for editing subpaths in a shape without having to explicitly select that layer.
> 
> >screen shot 2014-07-15 at 09 49 29
> 
> Pinging @mikeabdullah and @pieteromvlee 

Connect to BohemianCoding/Sketch#9169.


mikeabdullah:
Review Notes: Added a `BCPointValueOnAxis` function

mikeabdullah:
My bad, we had an equivalent function already.

PR #63, opened 2016-07-05
Code Review: Double-Buffer the MSTile's Rendering Contexts to Improve Performance (#9792)

randomsequence:
Code review for Double-Buffer the MSTile's Rendering Contexts to Improve Performance (#9792):

> Whilst profiling #9670 I noticed there's a lot of time spent doing copy-on-write at the start of the rending phase. This is happening because we're creating a `CGImageRef` from the `CGContextRef`, then using that as the backing for an on-screen layer. At this point the image and context both share the same memory. When we begin drawing to the context again that memory can no longer be shared, it has to be copied.
> 
> If we maintained two `CGContextRef`s for each tile, and flipped between them for alternate display / drawing phases we'd avoid that copy-on-write at no memory expense, since we're copying it anyway.

Connect to BohemianCoding/Sketch#9792.


randomsequence:
Review notes: this adds `NSDiscardableContext` to the object pool class. That gives the system an opportunity to empty the pool under memory pressure.

samdeane:
Reviewed: looks good. Also: different issue, same class, code re-use FTW :)

PR #64, opened 2016-07-06
Code Review: MSDocumentReader (#9731)

opsGavin:
Code review for MSDocumentReader (#9731):

> Part of #8880 - the idea is that this class replaces the current MSDocumentFile functionality for reading.

Connect to BohemianCoding/Sketch#9731.


Source/BCJSONDecoder.h
```diff
@@ -9,6 +9,10 @@
 /// NSKeyedUnarchiver etc work.
 @interface BCJSONDecoder : NSObject
 
+/// Helper method to set an error on the decoder. Mostly this will be done internally, but
+/// we may use it when (for instance) decoding external references.
+- (void)setErrorWithDescription:(nullable NSString*)description, ... NS_FORMAT_FUNCTION(1,2);
```
mikeabdullah: What does passing `nil` here do, clear out the error?

Source/BCJSONDecoder.h
```diff
@@ -9,6 +9,10 @@
...
+
 /// Decodes the contents of the file at \c url into an object tree.
 /// If an error occurs, this method will return nil with \c error filled in.
 + (nullable id)decodeFileAtURL:(nonnull NSURL*)url error:(NSError* __nullable __autoreleasing * __nullable)error;
```
mikeabdullah: Not direct to this pull, but annotating the whole file with `NS_ASSUME_NONNULL_BEGIN` avoids having to repeat ourselves so much.

Source/BCJSONDecoder.m
```diff
@@ -15,11 +15,12 @@ @interface BCJSONDecoder ()
 - (nonnull instancetype)initWithData:(NSData*)data;
 - (nonnull instancetype)initWithRootObject:(NSDictionary*)root;
 
-- (void)setErrorWithDescription:(nullable NSString*)description, ... NS_FORMAT_FUNCTION(1,2);
 @property (nonatomic, strong) NSError* error;
```
mikeabdullah: Generally, I'd say it's desirable to specify `copy` for the error.

Source/BCJSONDecoder.m
```diff
@@ -15,11 +15,12 @@ @interface BCJSONDecoder ()
...
-
 @property (nonatomic, strong) NSDictionary* root;
+@property (nonatomic, weak) NSData* data;
 @property (nonatomic, strong) NSDictionary* currentObjectDict;
+
+- (nullable id)decode;
```
mikeabdullah: Any chance of docs for this new property and method? I have no idea offhand what `data` is used for, or why it's `weak`.

Source/BCJSONDecoder.m
```diff
@@ -46,25 +47,41 @@ + (id)decodeFileAtURL:(NSURL*)url error:(NSError* __autoreleasing*)error {
...
 - (instancetype)initWithData:(NSData *)data {
+  self = [super init];
+  if (self) {
+    self.data = data;
+  }
+  return self;
+}
+
+- (void)setRoot:(NSDictionary*)root {
+  if (!root || [root isKindOfClass:[NSDictionary class]]) {
```
mikeabdullah: If something passes in a non-dictionary object here, would it be better to throw an exception? Generally I'd say that's better than quietly swallowing the mistake.

Source/BCJSONDecoder.m
```diff
@@ -46,25 +47,41 @@ + (id)decodeFileAtURL:(NSURL*)url error:(NSError* __autoreleasing*)error {
...
+  self.root = [NSJSONSerialization JSONObjectWithData:data options:0 error:&error];
   if (error) {
     self.error = error;
   }
-  return self;
+}
+
+- (NSData *)data {
+  NSAssert( nil, @"Shouldn't be called" );
+  return nil;
```
mikeabdullah: I find this particularly weird, why go to the trouble of declaring a property without a working getter?

Source/BCJSONDecoder.m
```diff
@@ -180,14 +201,16 @@ - (id)decode {
...
+  if (!self.error) { // One error could lead to others later - only record the first
+    if (description) {
+      va_list args;
+      va_start(args, description);
+      description = [[NSString alloc] initWithFormat:description arguments:args];
+      va_end(args);
+    }
 
-  self.error = [NSError errorWithDomain:@"BCJSONDecoder" code:0 userInfo:@{ @"decription" : description?description:@"Unknown error" }];
+    self.error = [NSError errorWithDomain:@"BCJSONDecoder" code:0 userInfo:@{ @"decription" : description?description:@"Unknown error" }];
```
mikeabdullah: The key `@"decription"` almost certainly has a typo. But more importantly, Apple provides `NSLocalizedDescriptionKey` for us, can we use it please.

mikeabdullah: I'd agree it's cleaner to do `if (!description) description = …` on a line above, rather than inline like this

mikeabdullah: Apple's general recommendation for custom error domains is they should use reverse domain notation. That said, is there actually any benefit to a custom domain here? We don't have any custom error codes yet, which is when they come in handy!

Source/BCJSONDecoder.h
opsGavin: It generates an error, but it's marked as an "Unknown Error" - could probably do with a comment to that effect...

Source/BCJSONEncoder.m
```diff
@@ -29,14 +29,21 @@ - (void)encodeAsJSON:(BCJSONEncoder*)coder {
...
   if (!exceptions) {
-     exceptions = @[@"NSMutableParagraphStyle",@"NSCTFontDescriptor",@"NSCachedRGBColor",@"NSConcreteAttributedString"];
+     exceptions = @[@"NSMutableParagraphStyle",
+                    @"NSCTFontDescriptor",
+                    @"NSCachedRGBColor",
+                    @"NSParagraphStyle",
+                    @"NSConcreteAttributedString",
+                    @"NSColorSpaceColor",
+                    @"NSCalibratedRGBColor",
+                    @"NSCachedWhiteColor"];
```
mikeabdullah: Listing raw private class names here strikes me as dodgy. Is there really no alternative?

Source/BCJSONEncoder.m
```diff
@@ -63,8 +70,13 @@ - (void)finishEncodingObject:(BCJSONEncoder*)coder {
   [coder writeData:"}" length:1];
 }
 
-- (NSString*)keyAttributeForObjectInJSON {
-  return [NSString stringWithFormat:@"%016llx", (uint64_t)self];
+- (void)encodeAsKey:(BCJSONEncoder*)coder {
+  // We're being asked to use a non-string or non-number value as a key in the dictionary.
+  // We'll use the object's address as the key, but this probably isn't going to
+  // be what we want and this situation probably needs to be addressed.
```
mikeabdullah: Yikes. I can't imagine when encoding the address would ever be a good idea. Suggest it's better to always fail with an exception.

Source/BCJSONEncoder.m
```diff
@@ -161,6 +180,11 @@ - (void)finishEncodingObject:(__unused BCJSONEncoder*)coder {
   // NOP
 }
 
+- (void)encodeAsKey:(BCJSONEncoder*)coder {
+  // Encoding a number as a key isn't valid JSON - encode as a string instead.
+  // Whoever is loading will need to sort this out at load time.
+  [coder encodeObject:self.stringValue];
```
mikeabdullah: On that basis, I suspect trying to be kind and accept numbers as input like this is just making the overall picture more complicated.

Source/BCJSONEncoder.m
```diff
@@ -98,20 +110,27 @@ - (void)finishEncodingObject:(BCJSONEncoder*)coder {
...
-     firstWriten = YES;
+    if (writeSeparator) {
+      [coder writeData:"," length:1];
+    } else {
+      writeSeparator = YES;
+    }
+
+    [key encodeAsKey:coder];
+    [coder writeData:":" length:1];
+    [coder encodeObject:value];
```
mikeabdullah: This bit surprises me. Do I understand this correctly, that dictionaries are encoded in a string form, rather than as dictionaries?

Source/BCJSONDecoder.h
opsGavin: Yes - that's a good point. Don't recall my reasoning for not doing it that way...

Source/BCJSONDecoder.m
```diff
@@ -116,10 +133,10 @@ - (id)decodeDictionary:(NSDictionary*)dict {
     if (base64archive) {
       // NSArchived data
       NSData* data = [[NSData alloc] initWithBase64EncodedString:base64archive options:0];
-      result = [NSUnarchiver unarchiveObjectWithData:data];
+      result = [NSKeyedUnarchiver unarchiveObjectWithData:data];
```
mikeabdullah: Good change

mikeabdullah:
Lots of questions but think I'm basically ok.

Source/BCJSONDecoder.m
opsGavin: Ha - yes definite typo. I didn't know about `NSLocalizedDescriptionKey` - good to know.

Source/BCJSONDecoder.m
opsGavin: I'm sure you're right, but curious to know your reasoning? You wouldn't expect an error to be mutated elsewhere after it's been created would you?

mikeabdullah: Indeed, `NSError`s are immutable. I consider them to be like other value objects (e.g. strings) where the generally accepted approach is to copy them.

Source/BCJSONDecoder.h
mikeabdullah: A comment would be much appreciated.

Source/BCJSONEncoder.m
opsGavin: It's really only a debugging precaution in case someone adds a new class which can't be archived properly. For these classes it's desirable to fall back to NSCoding, but if something else comes up we'd want to check it.

Open to suggestions though. One option would be to add categories to these classes and just assert here, but that seems more prone to error to me.

Source/BCJSONEncoder.m
opsGavin: I wavered about this. Failing means that the user will be stuck with a document that refuses to save - passing means we're going to end up with a problem in the document which we don't necessarily know anything about and which could cause strange issues further down the road.

I think on balance you're right - I'll update it.

Source/BCJSONEncoder.m
opsGavin: True - although I discovered this was actually happening :-/

Specifically in the `overrides` for symbol instances the keys are numbers. I've dealt with it by fixing up this dictionary in `MSImmutableSymbolInstance` but I do wonder if we wouldn't be better off asserting here and changing what gets written as well.

opsGavin: Yeah - you've convinced me. Think I'll change it 😉

Source/BCJSONEncoder.m
opsGavin: Well they're written out as a JSON dictionary `{ key : value, key2 : value2 }` - not sure if that's what you mean?

Source/BCJSONEncoder.m
mikeabdullah: I think I see. As a middle ground, why can't we operate on the `NSColor` class directly for example? I don't see a category as being necessary to achieve the desired effect.

Source/BCJSONEncoder.m
mikeabdullah: Aha, this code is lower level than I previously thought then, it's doing the bit of building raw output.

Source/BCJSONDecoder.h
```diff
@@ -5,43 +5,60 @@
...
 #import "BCJSONEncoder.h"
 
+NS_ASSUME_NONNULL_BEGIN
+
 /// The BCJSONDecoder class can be used to decode data in similar way to how
 /// NSKeyedUnarchiver etc work.
 @interface BCJSONDecoder : NSObject
 
+/// The \c userInfo property can be used by a consumer of this class to attach
+/// information pertenant to this current decoding.
```
pieteromvlee: 'pertinent', right?

Source/BCJSONEncoder.h
```diff
@@ -34,6 +34,10 @@
 /// Instead encoding should take place via encodeObject:to... methods.
 @interface BCJSONEncoder : NSObject
 
+/// The \c userInfo property can be used by a consumer of this class to attach
+/// information pertenant to this current decoding.
```
pieteromvlee: `pertinent`?

Source/BCJSONEncoder.m
```diff
@@ -63,8 +70,11 @@ - (void)finishEncodingObject:(BCJSONEncoder*)coder {
   [coder writeData:"}" length:1];
 }
 
-- (NSString*)keyAttributeForObjectInJSON {
-  return [NSString stringWithFormat:@"%016llx", (uint64_t)self];
+- (void)encodeAsKey:(BCJSONEncoder*)coder {
+  // We're being asked to use a non-string or non-number value as a key in the dictionary.
+  // We haven't got a mechinism to handle loading such a thing and the situation should be addressed.
```
pieteromvlee: 'mechanism'

Source/BCJSONEncoder.m
opsGavin: > As a middle ground, why can't we operate on the NSColor class directly for example?

Sorry - not sure what you mean here... my brain is probably still in holiday mode ;)

mikeabdullah: Well you've got a list of Apple's private concrete classes here, which they're entitled to change at any time. Instead have a list of `NSColor` etc, loop through, looking to see if the receiving object `isKindOfClass:NSColor` and so on.

PR #65, opened 2016-07-07
Code Review: Switch Copy / Paste to use JSON (#9853)

opsGavin:
Code review for Switch Copy / Paste to use JSON (#9853):

> Part of #8880. Copy and paste uses the archivers to generate pasteboard data. This should be switched to use the JSON archiver.

Connect to BohemianCoding/Sketch#9853.


opsGavin:
Review notes: Where we have a point, size or rect wrapped in an NSValue I'm now encoding that directly. It's possibly slightly quicker than encoding the raw data, but the main motivation is that it's more readability in the JSON.

pieteromvlee:
👍

Source/BCJSONDecoder.h
```diff
@@ -35,6 +35,10 @@ NS_ASSUME_NONNULL_BEGIN
 /// If an error occurs, this method will return nil with \c error filled in.
 - (nullable id)decodeData:(NSData*)data error:(NSError* __nullable __autoreleasing * __nullable)error;
 
+/// Parses an NSDictionary representation of JSON data into an object tree.
+/// If an error occurs, this method will return nil with \c error filled in.
+- (nullable id)decodeDictionary:(NSDictionary*)dict error:(NSError* __nullable __autoreleasing * __nullable)error;
```
pieteromvlee: Decoding a dictionary? Isn't a dictionary already a decoded representation of JSON rather than the other way round? Confused here.

Or does this go from NSDict > MSLayer and friends whereas the method above does NSData > NSDict > MSLayer behind the scenes?

opsGavin: Yes; behind the scenes we do NSData (which is basically the JSON string) -> NSDictionary -> MSLayer etc. This method just skips the first part.

Not sure if there's a better name that would express that more clearly?

For what it's worth, a future optimisation might be to parse the JSON directly to MSLayer etc, avoiding the intermediate NSDictionary. The NSDictionary route would still be useful in some cases, but this would prevent the overhead of creating the whole dictionary tree and storing it in memory. One for another day though... ;-)

PR #70, opened 2016-07-21
Code Review: Fix tiling problem in pixel zoom (#10022)

pieteromvlee:
Code review for Fix tiling problem in pixel zoom (#10022):

> Follow-up from #10008 where we fixed one problem but not the second. This is a ticket for the second problem:
> 
> 10008-zoom-level-pixel-bug

Connect to BohemianCoding/Sketch#10022.


pieteromvlee:
This is a bad idea! PR abanonded deliberately

PR #72, opened 2016-07-27
Code Review: Tidy up work for new JSON file format (#10085)

opsGavin:
Code review for Tidy up work for new JSON file format (#10085):

> The purpose of this issue is to consolidate the work on the new JSON file format to finish off #8880.
> 
> I started work on #9853 to use the JSON archiver for the copy and paste code, but after talking to @pieteromvlee it been decided to drop this from #8880. (It can still be done, but can be moved to v41 instead of v40).
> 
> There are a few, tidy up changes and optimisations in the #9853 branch which would be better in #8880, so those changes can be cherry-picked into this issue's branch.
> 
> I also need to check the QL plugin. At present it should work as before (reading the document and generating an image) but this needs checked. At some stage it would be desirable to embed a preview image in the zip archive, but we can safely leave this until v41 or later. This may not need much, but there are some linker errors building the Quicklook target which I need to look into.

Connect to BohemianCoding/Sketch#10085.


opsGavin:
Review notes: Just some optimisations. Rather than using `[NSString stringWithFormat...]` I'm dropping down to C / `printf`. It's not as tidy to read, but this stuff's called a lot during a save and it speeds thing up by about 30%

pieteromvlee:
30%?? That's quite a lot! 👍

PR #78, opened 2016-08-23
Code Review: Dragging from the LL should use promises (#4561)

opsGavin:
Code review for Dragging from the LL should use promises (#4561):

> The LL currently writes big files asynchronously, but not using promises. It probably should do.
> 
> See also #4991 - need to make sure we're reusing the maximum amount of code here.

Connect to BohemianCoding/Sketch#4561.


Source/NSURL+BCFoundation.h
```diff
@@ -0,0 +1,12 @@
+//
+//  Created by Gavin on 23/08/2016.
+//  Copyright © 2016 Bohemian Coding. All rights reserved.
+//
+
+@interface NSURL (BCFoundation)
+
+/// Returns a URL with any period stripped off the beginning of the file name.
+/// This method can be used to ensure that any file we're dealing with is not hidden.
+- (NSURL*)URLByStrippingInvisiblePeriod_bc;
```
pieteromvlee: I'm not a fan of these suffixes but more importantly `MSExportManager` has a similar method already:

`- (NSURL *)URLByStrippingInvisiblePeriodFromURL:(NSURL *)fileURL`

Can you update that class to use this new utility method?

opsGavin: Actually, `MSExportManager` is where it came from. I moved some code to write the exported data to a file from `MSExportManager` into `MSExporter` (Saving to a file didn't particularly seem the job of a manager and the move let me use it from one of the Sketch controller classes). There's still some other code in `MSExportManager` which still uses this method - so as I needed this in 2 places I thought it worth adding to BCFoundation.

pieteromvlee:
Oh I totally overlooked that you removed it there. How stupid of me. Off to QA then

PR #80, opened 2016-08-31
Code Review: Make Zipped JSON our official File Format (#8880)

samdeane:
Code review for Make Zipped JSON our official File Format (#8880):

> In #8756 we wrote a proof-of-concept to test the various approaches. Zipped JSON was surprisingly small compared to what we have now, and it didn't increase in size that much. The benefit of being able to do/allow server-side file parsing makes this the preferred option. Let's build it! :-D
> 
> This'll be for @opsGavin since he built the POC, but also pinging in @samdeane and @mikeabdullah since I know he shares a keen interest in all things zip.
> 
> For now I think it makes sense to only use the Intermediate Value Tree (#9190) for unarchiving since we win nothing by using it for archiving too. Yet we'll make a pluggable NSCoding-like class so we can easily do it later if needed
> ## 
> 
> Things to look out for:
> 1. We use `ZipKit` already so we should either use that or make sure we update existing code to deal with `minizip' (which the POC used). Either way, we should have 1 zip package
> 2. Make sure QuickLook keeps working with new and old files
> 3. Even though our file format might for the moment only be 1 json file, it's expected that we'll have other files in there at some point so if we start with a zipped directory from the start that might make sense
> ## 
> 
> Possible steps to take:
> - [x] Replace our MSArchiver with a custom serialiser class so we can plug in a JSON serialiser later #9365
> - [x] Replace our MSUnarchiver with a custom deserialiser so we can plug in a JSON deserialiser later #9402
> - [x] Write the above serialiser #9401
> - [x] Write a deserialiser than reads in a mutable value tree, can perform migrations on that and then pass that to the immutables in a custom MSUnarchiver-type subclass #9403
> - [x] Wrap the new JSON in a zipped folder with images and a `metadata.json` and make Sketch archive and unarchive that besides keeping support for unarchiving the old SQL format
> - [x] Adapt the QL plugin to check for this image and fall back to an SQL read for the old format
> - [ ] Make sure we can do migrations and detect documents that are too new like we can now
> - [ ] Make sure missing fonts etc are recognises like they used to be
> - [ ] Embed a preview of the first page in the Zip for QL and other tools
> - [ ] Make sure old versions of Sketch don't crash but display some kind of sensible error with encountering the new file
> - [ ] Make sure we can do some test migrations easily

Connect to BohemianCoding/Sketch#8880.


Source/BCJSONDataEncoder.m
```diff
@@ -0,0 +1,48 @@
...
+
+@implementation BCJSONDataEncoder
+- (instancetype)initWithData:(NSMutableData*)data {
+  self = [super init];
+  if (self) {
+    self.data = data;
+    if (!data) {
+      // Shouldn't be allowed here as data is marked as non-null, but reality probably
+      // dictates otherwise...
+      [self setErrorWithCode:0 description:@"Invalid output"];
```
samdeane: It would be good to actually define some error codes...

Source/BCJSONDataEncoder.h
```diff
@@ -0,0 +1,17 @@
+//
+//  Created by Gavin on 02/06/2016.
+//  Copyright © 2016 Bohemian Coding. All rights reserved.
+//
+
+#import "BCJSONEncoder.h"
+
+@interface BCJSONEncoder (BCJSONDataEncoder)
```
samdeane: Could do with header doc comments describing the purpose of this class (likewise elsewhere for the other classes).

samdeane: I'm not totally convinced of the need for the BCJSONFileEncoder/BCJSONDataEncoder distinction. Could you not just use an `NSOutputStream`? This seems to be reinventing the wheel...

Source/BCJSONDecoder.h
```diff
@@ -0,0 +1,68 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+/// The BCJSONDecoder class can be used to decode data in similar way to how
+/// NSKeyedUnarchiver etc work.
+@interface BCJSONDecoder : NSObject
+
+/// The \c userInfo property can be used by a consumer of this class to attach
+/// information pertinent to this current decoding.
+@property (nonatomic,weak,nullable) id userInfo;
```
samdeane: why weak?

Source/BCJSONDecoder.m
```diff
@@ -0,0 +1,247 @@
+//
+//  Created by Gavin on 03/06/2016.
+//  Copyright © 2016 Bohemian Coding. All rights reserved.
+//
+
+#import "BCJSONDecoder.h"
+
+// Defined in BCJSONEncoder.m
+extern NSString* BCJSONRootItemKey;
```
samdeane: Why are these exposed in this header if they're defined in BCJSONEncoder.m?

Source/BCJSONDecoder.m
```diff
@@ -0,0 +1,247 @@
...
+  NSMutableArray* result = [NSMutableArray array];
+  [array enumerateObjectsUsingBlock:^( NSObject* obj, NSUInteger idx __unused, BOOL* stop __unused) {
+     [result addObject:[self decodeObject:obj]];
+   }];
+  return result;
+}
+
+
+- (id)decodeObject:(NSObject*)object {
+  if ([object isKindOfClass:[NSDictionary class]]) {
```
samdeane: Can't you do this with a polymorphic method?

Source/BCJSONDecoder.m
```diff
@@ -0,0 +1,247 @@
...
+    if (description) {
+      va_list args;
+      va_start(args, description);
+      description = [[NSString alloc] initWithFormat:description arguments:args];
+      va_end(args);
+    } else {
+      description = @"Unknown error";
+    }
+
+    self.error = [NSError errorWithDomain:@"com.bohemiancoding.BCJSONDecoder" code:0 userInfo:@{ NSLocalizedDescriptionKey : description }];
```
samdeane: Error codes!

Source/BCJSONEncoder.m
```diff
@@ -0,0 +1,425 @@
+//  Created by Gavin on 02/06/2016.
+//  Copyright © 2016 Bohemian Coding. All rights reserved.
+
+#import "BCJSONEncoder.h"
+#import "BCJSONFileEncoder.h"
+#import "NSArray+BCFoundation.h"
+
+// Helper method to write directly to coder using a printf style format
+#define printfToData( coder, format, ... ) { char buffer[128]; buffer[127] = 0; \
```
samdeane: This ought to be an inline function.

samdeane: I can't help feeling that it could also just be a method on BCJSONEncoder, which would replace the call to `writeData` and thus not require any more method calls...

Source/BCJSONEncoder.m
```diff
@@ -0,0 +1,425 @@
...
+@end
+
+
+
+#pragma mark NSString + JSON
+@implementation NSString (BCJSONEncoder)
+
+- (NSString*)safeJSONValue {
+  NSString *result = self;
+  if ([self rangeOfCharacterFromSet:BCJSONAwkwardCharacters].location != NSNotFound) {
```
samdeane: this looks like it has the potential to be rather inefficient...

Source/BCJSONEncoder.m
```diff
@@ -0,0 +1,425 @@
...
+     exceptions = @[@"NSFontDescriptor",
+                    @"NSColor",
+                    @"NSParagraphStyle",
+                    @"NSAttributedString",
+                    @"NSColorSpaceColor"];
+    exceptions = [exceptions map:^id(NSString *className) {
+      return NSClassFromString(className);
+    }];
+  }
+  // Use NSAssert as BCFoundation doesn't know about ECCore
```
samdeane: The `ECAssert` macros are in `ECLogging`, which `BCFoundation` can see. Not that we *have* to use those macros, but...

Source/BCJSONEncoder.m
```diff
@@ -0,0 +1,425 @@
...
+  [coder encodeObject:[self safeJSONValue]];
+}
+@end
+
+
+#pragma mark NSDictionary + JSON
+@implementation NSDictionary (BCJSONEncoder)
+- (void)encodeAsJSON:(BCJSONEncoder*)coder {
+  __block BOOL writeSeparator = NO;
+  [self enumerateKeysAndObjectsUsingBlock:^(NSObject *key, NSObject *value, BOOL * stop __unused) {
```
samdeane: I think we might thank ourselves in the future if we encode dictionary keys in a deterministic order (eg sort them).

Source/BCJSONFileEncoder.m
```diff
@@ -0,0 +1,134 @@
...
+
+#pragma mark BCJSONFileEncoder
+
+@implementation BCJSONFileEncoder
+
+- (instancetype)initWithFile:(NSFileHandle*)output {
+  self = [super init];
+  if(self) {
+    self.output = output;
+    _buffer = malloc(BCJSONFileEncoderBufferSize);
```
samdeane: Why have you implemented your own level of buffering?

Source/BCJSONEncoder.m
opsGavin: The reason I did it as a define was that I found it to be slightly more optimal.

This is having the method as a static inline (highlighting the methods that use it):
screen shot 2016-08-31 at 10 31 22
Verses as a define:
screen shot 2016-08-31 at 10 31 36

This is a sizeable file and it's not a huge difference so I'm a little on the fence about whether the define is better than an inline.

Source/BCJSONDecoder.m
opsGavin: It's really only there because the decoder needs it as well. Maybe the decoder should declare it as extern at the top of the file?

Source/BCJSONEncoder.h
```diff
@@ -0,0 +1,72 @@
...
+@property (nonatomic,weak,nullable) id userInfo;
+
+/// Encodes the root object
+- (BOOL)encodeRootObject:(nullable NSObject*)root error:(NSError * __nullable __autoreleasing * __nullable)error;
+
+/// Encodes the passed in object and stores it as key
+- (void)encodeObject:(nullable NSObject*)object forKey:(nonnull NSString*)key;
+
+/// Encodes the passed in double value and stores it as key
+- (void)encodeDouble:(double)value forKey:(nonnull NSString*)key;
```
samdeane: For consistency really we should deal with foundation number types, which means CGFloat rather than double.

I always struggle a bit then with the idea of calling a method `Double` when it takes a type that is called `Float`.

For what it's worth, my solution in this case would be to call the methods `encodeReal:forKey:` and `decodeRealForKey:`.

Source/BCJSONDataEncoder.h
opsGavin: This was the direction I took to start, but I came to the conclusion that NSOutputStream wouldn't do what we needed. I'm struggling to remember what the issue was though - I'll need to look into it again to remind myself.

Source/BCJSONDecoder.h
opsGavin: Generally the consumer is going to be around for the duration of the write and can ensure that the `userInfo` is available while the write is happening. In fact it's quite likely that the consumer itself would be the `userInfo`. Making it strong would mean that the consumer would have to explicitly set it to `nil` again to avoid circular references.

Source/BCJSONDecoder.m
opsGavin: Yes - that would be better.

Source/BCJSONEncoder.h
opsGavin: The reason I've done it like this is that this is what `NSKeyedArchiver` does. We still need `NSKeyedArchiver` functionality in the model objects (or more correctly - we still need the `NSKeyedUnarchiver` methods). We could change all our methods to `encodeReal:` etc, but it would be inconsistent with Apple's stuff.

Source/BCJSONEncoder.m
opsGavin: Yeah - generally this accounts for around 6% of the save time.  I'm not sure of a better approach though TBH. `BCJSONAwkwardCharacters` is a static initialised at startup and I try to avoid calling this method when we can guarantee the string ourselves (e.g. I don't call it when writing keys as these are under our control, but it'll get called on user supplied values).

Suggestions very welcome! 😄

Source/BCJSONEncoder.m
opsGavin: Hmm - I'm not totally convinced. What sort of pain can you see us avoiding?

My worry is that this could be fairly inefficient. But, having said that, we probably don't encode that many dictionaries directly anyway.

Source/BCJSONFileEncoder.m
opsGavin: Neither `NSFileHandle` or the minizip writer (used in a class derived from this one) seem to do anything along these lines. Doing our own buffering makes a fairly significant difference.

Source/BCJSONDecoder.m
samdeane: I’d just put this in BCJSONEncoder.h, and make them explicitly public. If they are used by both BCJSONEncoder.m and BCJSONDecoder.m then they need to be shared, so having them public doesn’t seem unreasonable.

Source/BCJSONEncoder.m
samdeane: There’s something going wrong somewhere if the optimiser isn’t dissolving away an inline function. Are you profiling an optimised build?

Source/BCJSONEncoder.m
samdeane: The pain that I have already experienced a number of times when our tests start breaking (eg the dump tests) because the system changes the dictionary implementation and the keys come out in a different order. I think it may actually have happened again on 10.12.

Having different systems save the same file in slightly different ways doesn’t sound desirable to me. Imagine if the file is committed to version control, for example. Two users on different machines might cause the same file to be repeatedly marked as changed, when all they’ve actually done is auto-saved the file without modifying it.

Source/BCJSONFileEncoder.m
samdeane: Scary!

Source/BCJSONEncoder.h
samdeane: Yeah, fair enough.

I’m all for consistency with Apple, but NSKeyedArchiver dates back to Next days I think, and I’m sure that if Apple were starting now they’d not do things this way! For a start they’d probably have an actual foundation typedef for floating point numbers, rather than CGFloat!

PR #82, opened 2016-09-20
Code Review: Remove _root key from JSON (#10657)

opsGavin:
Code review for Remove _root key from JSON (#10657):

> Part of #8880 
> 
> With the current implementation all top level objects are encoded as `{"_root":object}`. We should just encode them as `object`

Connect to BohemianCoding/Sketch#10657.


pieteromvlee:
:+1:

PR #83, opened 2016-09-20
Code Review: Mechanism to migrate classes (#10661)

opsGavin:
Code review for Mechanism to migrate classes (#10661):

> Part of #8880 
> 
> There should be a mechanism for migrating renamed classes in the zipped JSON. See also the discussion in #10502

Connect to BohemianCoding/Sketch#10661.


Source/BCJSONDecoder.m
```diff
@@ -21,6 +23,46 @@ @interface BCJSONDecoder ()
...
 @implementation BCJSONDecoder
 
++ (void)setClass:(nullable Class)class forClassName:(NSString *)codedName {
+  if (!class) {
+    class = [NSNull class];
+  }
+  
+  if (!BCJSONDecoderClassMap) {
+    // Not thread safe, but we really don't expect to be called from anything other than the main thread at startup
+    ECAssertIsMainThread();
```
pieteromvlee: I do think its wrong to asset we're on the main thread. I can imagine a few scenarios where we want to load documents on background thread. Take the case of loading asset libraries for example.

I'd remove the assert and just specify in the header that the JSONDecoder isn't thread safe just as most of foundation isn't threadsafe

Source/BCJSONDecoder.m
```diff
@@ -21,6 +23,46 @@ @interface BCJSONDecoder ()
...
+    // Mapping onto ourselves - remove any previous mapping as there's no
+    // point spending time looking searching the dictionary for something that's
+    // going to give us default behaviour.
+    BCJSONDecoderClassMap[codedName] = nil;
+  } else {
+    BCJSONDecoderClassMap[codedName] = class;
+  }
+}
+
+- (Class)classToCreateForClassName:(NSString*)className {
```
pieteromvlee: `-classFromClassName` feels like a better name to me

Source/BCJSONDecoder.m
opsGavin: My thinking here was that this is a class method that we'd only expect to be called from one method at startup (e.g. in the `initialise` method of our serialiser class). I'd expect this to have been done by the time we start loading assert libraries and I'm not sure it would ever make sense to add additional mappings later on.

An alternative would be to do this in a `dispatch_once` block and synchronise any writing to the dictionary. That would guarantee the thread safety, but this approach always seems a little heavy handed to me if we're in control of when it gets called.

Source/BCJSONDecoder.m
opsGavin: 👍

Source/BCJSONDecoder.m
pieteromvlee: Ah you might be right there yes, ok. Ignore me then :)

pieteromvlee:
👍

PR #88, opened 2016-10-07
Code Review: Add support for layer exporting to the javascript API (#10671)

samdeane:
Code review for Add support for layer exporting to the javascript API (#10671):

> I think this would be the number one request many people have. In more concrete terms, it'd make Principle happy

Connect to BohemianCoding/Sketch#10671.


Tests/NSURL+BCFoundationTests.m
```diff
@@ -0,0 +1,69 @@
...
+}
+
+- (void)testFolderPath {
+  NSURL* base = [NSURL fileURLWithPath:@"/a/test/path"];
+
+  NSURL* url = [base URLByAppendingNameWithSeparators:@"subfolder/layer" defaultName:@"<untitled>"];
+  ECTestAssertIsEqual([url path], @"/a/test/path/subfolder/layer");
+}
+
+- (void)testCrazyShit {
```
opsGavin: As method names go - I guess this does what it says on the tin 👍 😛

samdeane: Hah - that might have been a previous version of me, but in this caseI just moved the code here from elsewhere...

PR #89, opened 2016-10-10
Code Review: Define error codes in JSON output (#10658)

opsGavin:
Code review for Define error codes in JSON output (#10658):

> Part of #8880
> 
> Where we use NSError we should define error codes to go along with them.

Connect to BohemianCoding/Sketch#10658.


Source/BCJSONEncoder.m
```diff
@@ -16,6 +16,14 @@
 NSString* BCJSONRawDataKey      = @"_data";
 NSCharacterSet *BCJSONAwkwardCharacters = nil;
 
+static NSString *BCJSONEncoderErrorDomain = @"com.bohemiancoding.BCJSONEncoder";
```
SheffieldKevin: and again a line spacing would be nice.

Source/BCJSONDecoder.m
```diff
@@ -234,20 +240,10 @@ - (id)decode:(NSError* __autoreleasing * )error {
...
-      description = [[NSString alloc] initWithFormat:description arguments:args];
-      va_end(args);
-    } else {
-      description = @"Unknown error";
-    }
-
-    self.error = [NSError errorWithDomain:@"com.bohemiancoding.BCJSONDecoder" code:0 userInfo:@{ NSLocalizedDescriptionKey : description }];
+- (void)setError:(NSError *)error {
+  if (!self.error) {
+    // One error could lead to others later - so we only record the first
```
opsGavin: That's right - it should be one shot.

Failing deep in the code will probably lead to other things failing further up the stack so the idea here is that we record the first, more pertinent, error we encounter rather than over-writing with a more generic error further on.

PR #90, opened 2016-10-26
Feature/10486

pieteromvlee:
Connect to https://github.com/BohemianCoding/Sketch/issues/10486


Source/NSArray+ConcurrentMap.h
```diff
@@ -0,0 +1,16 @@
+//  Created by Pieter Omvlee on 26/10/2016.
+//  Copyright © 2016 Bohemian Coding. All rights reserved.
+
+#import "NSArray+BCFoundation.h"
+
+typedef void(^BCConcurrentMapCompletionBlock)(NSArray *mappedArray);
```
randomsequence: Could `mappedArray` be `NULL`?

Source/NSArray+ConcurrentMap.h
```diff
@@ -0,0 +1,16 @@
...
+
+@interface NSArray (ConcurrentMap)
+/** A parallel version of the regular \c map: function using NSOperation under the hood to parallelise as best as possible
+ @param maxConcurrentCount the number of concurrent operations allowed. For example, for data-bandwidth-heavy operations you can set this to the number of available cores instead of 0, which means unbounded
+ @param completionBlock will be called when the concurrent operation is done. The completionBlock will be called on a background queue, make no guarantees which queue it is on
+ */
+- (void)concurrentMap:(BCMapBlock)block
+   maxConcurrentCount:(NSInteger)maxConcurrentCount
+      completionBlock:(BCConcurrentMapCompletionBlock)completionBlock;
```
randomsequence: Can `completionBlock` be `NULL`?

Source/NSArray+ConcurrentMap.h
```diff
@@ -0,0 +1,16 @@
...
+
+typedef void(^BCConcurrentMapCompletionBlock)(NSArray *mappedArray);
+
+@interface NSArray (ConcurrentMap)
+/** A parallel version of the regular \c map: function using NSOperation under the hood to parallelise as best as possible
+ @param maxConcurrentCount the number of concurrent operations allowed. For example, for data-bandwidth-heavy operations you can set this to the number of available cores instead of 0, which means unbounded
+ @param completionBlock will be called when the concurrent operation is done. The completionBlock will be called on a background queue, make no guarantees which queue it is on
+ */
+- (void)concurrentMap:(BCMapBlock)block
+   maxConcurrentCount:(NSInteger)maxConcurrentCount
```
randomsequence: See previous comment about using some options rather than an explicit concurrent count.

Source/NSArray+ConcurrentMap.h
```diff
@@ -0,0 +1,16 @@
...
+#import "NSArray+BCFoundation.h"
+
+typedef void(^BCConcurrentMapCompletionBlock)(NSArray *mappedArray);
+
+@interface NSArray (ConcurrentMap)
+/** A parallel version of the regular \c map: function using NSOperation under the hood to parallelise as best as possible
+ @param maxConcurrentCount the number of concurrent operations allowed. For example, for data-bandwidth-heavy operations you can set this to the number of available cores instead of 0, which means unbounded
+ @param completionBlock will be called when the concurrent operation is done. The completionBlock will be called on a background queue, make no guarantees which queue it is on
+ */
+- (void)concurrentMap:(BCMapBlock)block
```
randomsequence: Wouldn't make much sense, but can `block` be `NULL`?

Tests/ConcurrentMapTest.m
```diff
@@ -0,0 +1,41 @@
...
+      if (i==199)
+        NSLog(@"!");
+    }
+    
+    //make sure the completion block gets called
+    [expectation fulfill];
+  }];
+  
+  //this shouldn't take too long
+  [self waitForExpectationsWithTimeout:1.0 handler:nil];
```
samdeane: If this is a long-running test (eg > 10ms), you should make sure it's not enabled in the Sketch scheme. It's should be enabled in the Sketch Jenkins (Integration) scheme of course.

Source/NSArray+ConcurrentMap.h
```diff
@@ -0,0 +1,20 @@
...
+#import "NSArray+BCFoundation.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+typedef void(^BCConcurrentMapCompletionBlock)(NSArray *mappedArray);
+
+@interface NSArray (ConcurrentMap)
+/** A parallel version of the regular \c map: function using NSOperation under the hood to parallelise as best as possible
+ @param maxConcurrentCount the number of concurrent operations allowed. For example, for data-bandwidth-heavy operations you can set this to the number of available cores instead of 0, which means unbounded
+ @param completionBlock will be called when the concurrent operation is done. The completionBlock will be called on a background queue, make no guarantees which queue it is on
```
mikeabdullah: So my _expectation_ of a method like this is it would block until mapping is finished. If you wanted to avoid blocking the main thread, the answer would be to execute _it_ on a worker queue.

But as I understand it, that's not the case here? This method schedules all the mapping work on worker queues, and then immediately returns. A `completionBlock` is used to know when the work is finished. I'm not saying that's wrong, but could we document it as such please?

Source/NSArray+ConcurrentMap.m
```diff
@@ -0,0 +1,50 @@
...
+  //the final operation maps from the operations to a result array;
+  NSBlockOperation *finalOperation = [NSBlockOperation blockOperationWithBlock:^{
+    NSArray *result = [operations map:^id(BCConcurrentMapOperation *op) {
+      return op.result;
+    }];
+    completionBlock(result);
+  }];
+  
+  //submit all operations and add the finalOperation as dependency to all
+  //we can be sure that the concurrent operations are all finished and wil all have a result set by this time
```
mikeabdullah: tiny typo "wil" -> "will"

Source/NSArray+ConcurrentMap.m
```diff
@@ -0,0 +1,50 @@
...
+  for (NSOperation *operation in operations) {
+    [queue addOperation:operation];
+    [finalOperation addDependency:operation];
```
mikeabdullah: In my experience, having a lot of dependencies can quickly deteriorate performance a lot. I think Apple have improved that somewhat in recent OS releases, but not 100% sure.

If we made this method synchronous, you could just do `[queue waitUntilAllOperationsAreCompleted]`. Food for thought.

Source/NSArray+ConcurrentMap.m
```diff
@@ -0,0 +1,50 @@
...
+}
+@end
+
+@implementation NSArray (ConcurrentMap)
+
+- (void)concurrentMap:(BCMapBlock)block maxConcurrentCount:(NSInteger)maxConcurrentCount completionBlock:(BCConcurrentMapCompletionBlock)completionBlock {
+  
+  NSOperationQueue *queue = [NSOperationQueue new];
+  if (maxConcurrentCount > 0)
+    queue.maxConcurrentOperationCount = maxConcurrentCount;
```
mikeabdullah: Braces please

Source/NSArray+ConcurrentMap.h
mikeabdullah: I agree with the current code; there's no value in passing `NULL`

Source/NSArray+ConcurrentMap.h
mikeabdullah: I also think this is fine. We can ease up the requirement to accept `NULL` if it ever becomes desirable.

Source/NSArray+ConcurrentMap.h
```diff
@@ -0,0 +1,20 @@
...
+ */
+- (void)concurrentMap:(BCMapBlock)block
+   maxConcurrentCount:(NSInteger)maxConcurrentCount
+      completionBlock:(BCConcurrentMapCompletionBlock)completionBlock;
+@end
+
+NS_ASSUME_NONNULL_END
```
mikeabdullah: My biggest concern now is discoverability. We have two existing mapping methods in `NSArray+BCFoundation`. Looking there, I'd have no idea a parallelised/async variant exists elsewhere in the codebase. Perhaps it's best saved for a future ticket, but I think we should do some cleanup:

- Ditch the separation of `map:` and `mapWithIndex:` variants. Always pass in the index, like Apple does. There's no harm
- A slightly more Cocoa like name would be good IMO: `mapUsingBlock:`
- Pull them into a `NSArray (Map)` category so are all in one place. To keep it clean, that could be declared inside of `NSArray+BCFoundation.h`, but implemented in its own file

mikeabdullah: and actually on that note, perhaps take the opportunity now to pass in the index to this new method too, even if we aren't using it at present.

mikeabdullah: In fact, come to think of it, for cleaner client code of this method, how about reordering the arguments slightly:

`-mapWithMaxConcurrencyCount:usingBlock:completionBlock:`

Source/NSArray+ConcurrentMap.m
pieteromvlee: Our tile rendering operations rely heavily on Operation dependencies so if thats the case that might be a bigger problem...

Source/NSArray+ConcurrentMap.h
pieteromvlee: Indeed that is not the case. Maybe a rename is in order?

Source/NSArray+ConcurrentMap.m
mikeabdullah: Indeed. Best leave it be for now probably then.

Source/NSArray+ConcurrentMap.h
mikeabdullah: I think the name is basically fine (although as I said, reordering arguments would be great), and that I wasn't absorbing it properly in the diff when intermingled with comments. I think the header file makes it clearer.

Source/NSArray+ConcurrentMap.h
pieteromvlee: I like that 👍

randomsequence: No love for [my earlier comments](https://github.com/BohemianCoding/BCFoundation/pull/90#pullrequestreview-6015713) about concurrency count vs QoS?

Source/NSArray+ConcurrentMap.h
```diff
@@ -12,9 +12,9 @@ typedef void(^BCConcurrentMapCompletionBlock)(NSArray *mappedArray);
  @param maxConcurrentCount the number of concurrent operations allowed. For example, for data-bandwidth-heavy operations you can set this to the number of available cores instead of 0, which means unbounded
  @param completionBlock will be called when the concurrent operation is done. The completionBlock will be called on a background queue, make no guarantees which queue it is on
  */
-- (void)concurrentMap:(BCMapBlock)block
-   maxConcurrentCount:(NSInteger)maxConcurrentCount
-      completionBlock:(BCConcurrentMapCompletionBlock)completionBlock;
+- (void)mapWithMaxConcurrencyCount:(NSInteger)maxConcurrentCount
+                        usingBlock:(BCMapBlock)block
+                   completionBlock:(BCConcurrentMapCompletionBlock)completionBlock;
```
mikeabdullah: Je t'aime

Source/NSArray+ConcurrentMap.h
mikeabdullah: Hmmmm. I've not looked in QoS all that much to date.

Reading through some of Apple's docs on the subject, I'd say that in our particular case of resizing images, we should be using `NSQualityOfServiceUserInitiated`. I strongly think it's not `BCFoundation`'s place to decide on that, so it would make the most sense to have an extra argument for passing in the QoS.

To my understanding, QoS is mostly about prioritising system resources between apps, and managing battery life. It doesn't solve the problem where given a large number of non-CPU-constrained tasks, GCD will attempt to run them all at once, wasting effort. So that suggests to me we still want to pass a limit.

That leaves us with a method that takes four arguments, something like `mapWithMaxConcurrencyCount:qualityOfService:usingBlock:completionBlock:`
It feels to me like it would be cleaner if we ditched the completion block and returned the result synchronously. Yes, it'd be a tiny tiny bit less efficient, but more readable/normal.

randomsequence: I was thinking it would look more like:

```objc
typedef NS_OPTIONS(NSUInteger, BCMapOptions) {
  BCMapConcurrencyLimitDefault = 0,
  BCMapConcurrencyLimitNumberOfCores = 1<<0,
  
  BCMapQoSDefault = 1<<8,
  BCMapQoSUserInitiated = 1<<9,
  …
  
  BCMapDefaultOptions = BCMapConcurrencyLimitDefault | BCMapQoSDefault
};


- (void)mapWithOptions:(BCMapOptions)options
            usingBlock:(BCMapBlock)block
       completionBlock:(BCConcurrentMapCompletionBlock)completionBlock;
```

…but in typing that I realise options aren't an ideal match here.

mikeabdullah: I wouldn't say it's too far out, Cocoa uses option parameters like that in some places.

I'm wondering, since the QoS concept was introduced, have we ever gone back and inspected our usage throughout the code? I'm thinking it could be good to do an audit of all uses of `NSOperation(Queue)` to see if they've an appropriate QoS assigned. And if so, that could be a good time to deal with this further.

samdeane: > I'm wondering, since the QoS concept was introduced, have we ever gone back and inspected our usage throughout the code?

No, we absolutely haven’t, and it would be a good idea I think. It’s that kind of stuff that I fear gets neglected through lack of there being a clear person with a role focussed on those sorts of cross-module things.

I have a vague recollection from the activity tracking stuff that QoS is interesting in that it might be “sticky” - in the sense that if you do something which creates a new process/thread/task, it might inherit the existing QoS (I may just be thinking of the activity, which definitely works like that, but I think the QoS stuff was introduced at the same time and may be related).

samdeane: Incidentally I’m not in favour of the blocking approach, although I understand why @mikeabdullah is proposing it.

If you consider how `NSOperation` / `gcd` works under the hood, then I think blocking a thread is never likely to be optimal. 

I’m pretty sure that neither of them do fancy time-slicing _within_ their own worker threads. What that means is that if you block inside an operation with some sort of synchronous wait, that thread really is frozen until the wait completes. Which means that the system is forced to spawn a new worker thread to replace the one that you’ve just blocked, so that they maintain a pool and make efficient use of all cores.

`NSOperation`/`gcd` is smart enough to manage this behind the scenes, but I don’t think that it’s as efficient as being able to queue up an additional block, which will consume zero resources until a kernel-level notification happens to say that the synchronous thing is done.

samdeane: > It doesn't solve the problem where given a large number of non-CPU-constrained tasks, GCD will attempt to run them all at once, wasting effort. So that suggests to me we still want to pass a limit.

Do you have proof that this happens, because my understanding is absolutely that `gcd` shouldn't do that. It should work through the tasks a few at a time, picking a sensible level of concurrency to fit the hardware.

If you’ve seen this happening in the past, I wonder if it’s actually a symptom of something else. I can see how having a ton of _really small_ tasks might cause some unacceptable overhead, since the effort of queuing/dequeuing might start to become significant. Something is definitely wrong though if it’s happening with long-running tasks.

samdeane: Similarly a large number of tasks which are _blocking_ internally would be bad - for the very reason I gave earlier - which is why it’s not a great idea in general to block inside a `gcd` task or an `NSOperation`.

samdeane: (as an aside, presumably for resizing images we _are_ cpu constrained? I’d certainly expect so, unless you mean that we’re constrained by memory io instead? Which is a whole other can of worms...)

pieteromvlee: I tried not having a limit and that was decidedly worse; GCD is not smart enough to deal with this. I've dealt with many variations of this in the past, and so has @mikeabdullah. You can absolutely get worse performance when the job isn't strictly CPU bound. The limit is necessary.

Not a fan of the blocking though, which is why I agree with @samdeane

mikeabdullah: Chiming in late, sorry.

In my experience things like memory io definitely become a bottleneck for image handling. The moment one of its worker threads is waiting on that, GCD thinks "oh look, spare CPU!" and dequeues the next task. Which then happens to _also_ wait on memory IO, and you end up thrashing wildly.

I ran into this with Sandvox and image scaling and just went with a simple queue limited to 2 ops at a time. Pieter tried unbounded with our example document for this ticket and found it to be noticeably slower. I don't think there was any further experimentation, but this is fine by me.

PR #104, opened 2017-02-16
Code Review: String encoding bug in New File Format (#12499)

opsGavin:
Code review for String encoding bug in New File Format (#12499):

> I heard this evening from Tim van Damme (AbstractApp) that they found a string encoding bug in our new file format. Something with emoji's skin color not being saved properly perhaps?
> 
> I hear that they sent @samdeane an email with more details. And also ping @opsGavin of course. And @BohemianCoding/qa 

Connect to BohemianCoding/Sketch#12499.


opsGavin:
Review Notes: If we're trying to encode a string with can't be converted to UTF8 we now drop back to encoding the string using `NSKeyedArchiver` as base64.

An alternative would be to encode the raw bytes as base64 along with the encoding. I think that would produce cleaner (more portable) JSON and I was quite tempted to take that approach. But I figured the `NSKeyedArchiver` approach might be safer. It would be good to get a second opinion on that though.

mikeabdullah:
I'm trying to wrap my head around this. UTF-8 is supposed to be able to represent anything within the unicode spec. Which NSString is supposed to conform to. Yet somehow we're getting an NSString into the app which can't be converted to UTF-8.

I suppose it's perfectly possible to arrive at such a string when it's created by referring to some raw data or something like that.

But then I wonder is trying to preserve such strings using keyed archiving really the best thing for us to do? Can we do something like `-[NSString dataUsingEncoding:allowLossyConversion:]` instead to ditch the invalid characters?

opsGavin:
I'm on the fence about that one @mikeabdullah. On the one hand the string is obviously fairly knackered (to use the technical parlance) and it displays inconsistently within Sketch: 

screen shot 2017-02-16 at 16 18 33

So cleaning it up might be good. But on the other hand if we throw away data the user thinks is valid are we going to be criticised for that?

What do other @BohemianCoding/coders think?

samdeane:
I think I'm with @opsGavin on this one. 

I'm guessing that an invalid string can be placed onto the clipboard by something else, and then we'll gamely paste it in? That's not to mention the havoc that plugins can cause. Preserving the actual bytes is probably safest.

PR #106, opened 2017-02-26
Code Review: Implement generic download manager for plugin updating (#12065)

SheffieldKevin:
Code review for Implement generic download manager for plugin updating (#12065):

> Part of epic #12035 
> 
> Now that we need to download plugins we should probably write some kind of download manager. Later on we can change other parts of the app that download things to use the same class, but lets start with plugins.
> 
> Naturally the download manager should take care of asynchronously downloading content with something like a callback block when completed. Detailed progress into is not necessary yet since we're not yet downloading things that are super big.
> 
> Even though it'll just be used for plugins for now it should of course be able to download anything, nor make any assumptions about the API being tailored to plugins.
> 
> Also; tests.

Connect to BohemianCoding/Sketch#12065.


Source/BCDownloadManager.h
```diff
@@ -0,0 +1,102 @@
...
+extern NSString *BCDownloadManagerDownloadTaskState;
+
+@protocol BCDownloadManagerDelegate;
+
+/**
+ A simple download manager.
+ */
+@interface BCDownloadManager <NSURLSessionDelegate> : NSObject
+
+/// The plugin manager identifier.
```
SheffieldKevin: The identifier can probably go.

Source/BCDownloadManager.h
```diff
@@ -0,0 +1,102 @@
...
+@protocol BCDownloadManagerDelegate
+
+/// All protocol methods are optional for the case where we use the blocks approach.
+@optional
+
+/// The status of one of the downloads has changed.
+- (void)downloadManager:(BCDownloadManager *)downloadManager downloadTaskHasUpdatedForIdentifier:(NSString *)identifier;
+
+/// The number or order of the tasks may have changed. If you have a table view you might want to reload data.
+- (void)downloadManagerTasksUpdated:(BCDownloadManager *)downloadManager;
```
SheffieldKevin: That looks like an oversight. I'll need to update that.

Source/BCDownloadManager.h
```diff
@@ -0,0 +1,102 @@
...
+
+/// The file in \c downloadedURL in its implementation of this delegate method should be moved or copied from the downloadedURL in this method.
+- (void)downloadManager:(BCDownloadManager *)downloadManager downloadTaskFinishedForIdentifier:(NSString *)identifier downloadedURL:(NSURL *)downloadedURL;
+
+/// Delegate method called for dealing with download errors.
+- (void)downloadManager:(BCDownloadManager *)downloadManager downloadTaskFailedWithError:(NSError *)error forTaskIdentifier:(NSString *)identifier;
+
+/// Delegate method called when there is a missing file to download.
+- (void)downloadManager:(BCDownloadManager *)downloadManager downloadTaskDestinationRequestDoesNotExistForIndex:(NSInteger)taskIndex URL:(NSURL *)url;
```
SheffieldKevin: and again.

Source/BCDownloadManager.m
```diff
@@ -0,0 +1,233 @@
...
+NSString *BCDownloadManagerNumberOfBytesDownloaded = @"BCDownloadManagerTestNumberOfBytesDownloaded";
+NSString *BCDownloadManagerDownloadTaskState = @"BCDownloadManagerDownloadTaskState";
+
+@interface BCDownloadManager ()
+
+/// The NSURLSession for downloading that will deal with redirections etc. for us.
+@property (nonatomic, strong) NSURLSession *session;
+
+/// The completion handler for a background session.
+// @property (nonatomic, strong) void(^completionHandler)();
```
SheffieldKevin: I'll leave this. Since I'm now trying to use the download manager in my next task it will expose whether it is likely a download manager completion task is needed or not. If it doesn't prove to be so I'll get rid of it then.

Source/BCDownloadManager.m
```diff
@@ -0,0 +1,233 @@
...
+@implementation BCDownloadManager
+
+- (instancetype)initWithIdentifier:(NSString *)identifier
+              sessionConfiguration:(NSURLSessionConfiguration *)sessionConfiguration
+                          delegate:(id <BCDownloadManagerDelegate>)delegate {
+//                 completionHandler:(void(^)())completionHandler {
+  if ((self = [super init])) {
+    _canAcceptNewDownloadTasks = YES;
+    _hasBeenInvalidated = NO;
+    _identifier = identifier;
```
SheffieldKevin: The download manager itself at this point is not used anywhere in sketch. The only place the download manager code is used is in tests. I was imagining a possibility where we might want to send a notification that a file has been downloaded. In which case knowing if this download manager is responsible for downloading appcasts, sketchplugin updates or something else would be useful. It may be overkill.

Source/BCDownloadManager.h
```diff
@@ -0,0 +1,102 @@
...
+@property (nonatomic, readonly) BOOL canAcceptNewDownloadTasks;
+
+/// Number of processing download tasks.
+@property (nonatomic, readonly) NSInteger numberOfUnfinishedDownloadTasks;
+
+/// Identifiers for all download tasks.
+@property (nonatomic, readonly) NSArray <NSString*> *allTasksIdentifiers;
+
+- (instancetype)initWithIdentifier:(NSString *)identifier
+              sessionConfiguration:(NSURLSessionConfiguration *)sessionConfiguration
```
SheffieldKevin: I started with assuming the background session configuration was the one we would want by default. But that proved to be untestable as it downloads on a different process to the application making it impossible to mock. To write tests I could either replace the default choice or provide an init method that allows us to provide the download session configuration. I decided on the later approach because I thought that being able to use the background configuration was also a reasonable choice for downloading.

For example if a sketch plugin writers site was hacked and the plugin replaced with a dodgy download aimed at crashing the download process, instead of sketch crashing only the background download process would crash.

opsGavin: Makes sense. It's always a little nasty having to define the interface to fit the tests, but sometime it just can't be helped.

Source/BCDownloadManager.h
```diff
@@ -0,0 +1,106 @@
...
+ @param delegate Not needed if all download tasks added to the download manager take a completion handler.
+ @param completionHandler Called if the configuration is a background download or if finishTasksAndInvalidate is called when all tasks have finished.
+ */
+
+- (instancetype)initWithSessionConfiguration:(NSURLSessionConfiguration *)sessionConfiguration
+                                    delegate:(nullable id<BCDownloadManagerDelegate>)delegate
+                           completionHandler:(nullable void (^)())completionHandler;
+
+/// Add a download task by creating a default request object for the download URL. You should implement the delegate methods if creating download tasks with this method.
+- (nullable NSURLSessionDownloadTask *)addDownloadTaskWithIdentifier:(NSString *)identifier downloadRequestURL:(NSURL *)downloadRequestURL;
```
opsGavin: Discussed this in Slack, but just for the record. I think this would be better allocating the identifier itself (so that we can guarantee it's unique) and returning that. I don't think the outside world needs to know about the `NSURLSessionDownloadTask` - but if it really had to we could add a `downloadTaskForIdentifier:` method.

For some reason my original comment on this got lost in the either - sorry @SheffieldKevin  😞

SheffieldKevin:
Thanks Gavin, I appreciated your input

PR #110, opened 2017-03-10
Code Review: Implement plugin manager API - Installing Updates (#12067)

SheffieldKevin:
Code review for Implement plugin manager API - Installing Updates (#12067):

Connect to BohemianCoding/Sketch#12067.


Source/BCDownloadManager.m
```diff
@@ -18,7 +18,8 @@ @interface BCDownloadManager ()
 /// The dictionary containing the currently active tasks. Tasks are removed when they have completed.
 @property (nonatomic, strong) NSMutableDictionary <NSString *, NSURLSessionDownloadTask *> *downloadingTasks;
 
-@property (nonatomic, strong) id <BCDownloadManagerDelegate> delegate;
+/// The download manager delegate.
+@property (nonatomic, weak) id <BCDownloadManagerDelegate> delegate;
```
mikeabdullah: Are you sure we want to do this? The API of `BCDownloadManager` looks to me like it's closely modelled on `NSURLSession`. `NSURLSession` has a strong delegate because of the asynchronous way it operates. There's no need to think about what happens if the delegate goes away or changes during the operation. Instead it's fixed for the lifetime of the op.

Once all ops are over and the session/manager invalidated, then the delegate is dropped, breaking any retain cycle.

Source/BCDownloadManager.m
```diff
@@ -177,7 +178,8 @@ - (NSString *)identifierForDownloadTask:(NSURLSessionDownloadTask *)downloadTask
 }
 
 - (BOOL)delegateRespondsToSelector:(SEL)selector {
-  if (self.delegate && [(NSObject *)(self.delegate) respondsToSelector:selector]) {
+  id <BCDownloadManagerDelegate> strongDelegate = self.delegate;
+  if (strongDelegate && [(NSObject *)(strongDelegate) respondsToSelector:selector]) {
```
mikeabdullah: I'm pretty sure we can do a cleaner job than all this casting. If `BCDownloadManagerDelegate` conforms to `NSObject`, problem solved.

Source/BCDownloadManager.m
SheffieldKevin: I initially had it as a strong reference, but the static analyser complained. I'll have a look again.

Source/BCDownloadManager.m
SheffieldKevin: 👍

Source/BCDownloadManager.m
```diff
@@ -18,6 +18,8 @@ @interface BCDownloadManager ()
 /// The dictionary containing the currently active tasks. Tasks are removed when they have completed.
 @property (nonatomic, strong) NSMutableDictionary <NSString *, NSURLSessionDownloadTask *> *downloadingTasks;
 
+/// The download manager delegate.
+// @property (nonatomic, weak) id <BCDownloadManagerDelegate> delegate;
 @property (nonatomic, strong) id <BCDownloadManagerDelegate> delegate;
```
mikeabdullah: I can't help thinking this repeated content is simply going to confuse anyone reading the code in future. Surely we can do something better?

Source/BCDownloadManager.h
```diff
@@ -83,7 +83,7 @@ extern NSString *BCDownloadManagerDownloadTaskState;
 
 #pragma mark BCDownloadManagerDelegate
 
-@protocol BCDownloadManagerDelegate
+@protocol BCDownloadManagerDelegate <NSObject>
```
mikeabdullah: 👍 thanks

Source/BCDownloadManager.m
```diff
@@ -177,7 +179,8 @@ - (NSString *)identifierForDownloadTask:(NSURLSessionDownloadTask *)downloadTask
 }
 
 - (BOOL)delegateRespondsToSelector:(SEL)selector {
-  if (self.delegate && [(NSObject *)(self.delegate) respondsToSelector:selector]) {
+  id <BCDownloadManagerDelegate> strongDelegate = self.delegate;
+  if (strongDelegate && [strongDelegate respondsToSelector:selector]) {
```
mikeabdullah: Personally I think it's neater to simply collapse this down to a single check of:

```if ([self.delegate responds…```

PR #116, opened 2017-04-06
Code Review: Corrupted files with the new file format (#13110)

opsGavin:
Code review for Corrupted files with the new file format (#13110):

Connect to BohemianCoding/Sketch#13110.


opsGavin:
This isn't a pretty fix and maybe we can tidy later but...

Basically the issue is that the JSON had invalid ascii characters (backspaces, end of txt, etc). If Apple's JSON decoder fails we try to look for these characters in the raw data, strip them out and try again.

PR #119, opened 2017-04-12
Code Review: Layers larger than parent Artboard don't resize correctly within a Symbol Instance (#11589)

opsGavin:
Code review for Layers larger than parent Artboard don't resize correctly within a Symbol Instance (#11589):

Connect to BohemianCoding/Sketch#11589.


opsGavin:
Review Notes: This turned into a reasonably sizeable change. Basically the previous logic was that the returned rectangle was expected to have to fit into the destination size. However, this wasn't really the right thing to be doing and the change is that such constraints have been removed.

PR #134, opened 2017-07-20
Code Review: Don't fail to open corrupt files if it can be avoided. (#13803)

opsGavin:
Code review for Don't fail to open corrupt files if it can be avoided. (#13803):

Connect to BohemianCoding/Sketch#13803.


Source/BCJSONDecoder.m
```diff
@@ -281,6 +278,8 @@ - (id)decode:(NSError* __autoreleasing * )error {
 }
 
 - (void)setError:(NSError *)error {
+  // Log the error so that we can tell the user to look in the system log for more info about the failure
```
SheffieldKevin: I thought we had ECLogging available in BCFoundation. Would it be better to use that than NSLog here.

opsGavin: I wasn't sure if ECLogging had an "always log" method?

PR #136, opened 2017-07-28
Code Review: new targets for BCFoundation - proof of concept (#14308)

samdeane:
Code review for “new targets for BCFoundation - proof of concept” (#14308):

Connect to BohemianCoding/Sketch#14308.


Source/BCGeometry.h
```diff
@@ -110,57 +111,18 @@ CGSize BCUnionSize(CGSize sizeA, CGSize sizeB);
...
-  BCConstraintMinYSizeable   = 1U << 5,
-  
-  // If it's really desirable to have everything fixed this will do it.
-  BCConstraintAllFixed       = 1U << 6,
+#pragma mark - Strings
 
-  BCConstraintAllSizeable    =  BCConstraintMinXSizeable | BCConstraintWidthSizeable  | BCConstraintMaxXSizeable |
-                                BCConstraintMinYSizeable | BCConstraintHeightSizeable | BCConstraintMaxYSizeable
-};
+NSString* BCStringFromPoint(CGPoint point);
```
randomsequence: Should we have the inverse of these functions too? (`BCPointFromString` etc)

Source/BCGeometry.m
```diff
@@ -258,137 +259,46 @@ BOOL BCRangeContainsRange(NSRange outerRange, NSRange innerRange) {
...
-      break;
-  }
-  
-  *pMinVal = minVal;
-  *pMaxVal = maxVal;
+
+#if TARGET_OS_IOS
+
+NSString* BCStringFromPoint(CGPoint point) {
+  return NSStringFromCGPoint(point);
```
randomsequence: I don't think there's any guarantee that `NSStringFromCGPoint()` (iOS) produces the same string as `NSStringFromPoint()` (macOS). If we were to store those strings in the document, or use them to obtain a hash for example, we might come unstuck.

I would either add a note in the header suggesting that the string might vary depending on platform.

Source/BCGeometry.h
```diff
@@ -110,57 +111,18 @@ CGSize BCUnionSize(CGSize sizeA, CGSize sizeB);
...
- Depending on the constraints the width and/or height could be zero.
- */
-CGSize BCMinumumSizeForRectWithConstraints( CGSize containerSize, CGRect inner, BCConstraint constraints );
+#if !TARGET_OS_IOS
 
 /**
  Returns an NSEdgeInsets with the left right top bottom sides calculated. It is assumed that the child rect falls fully inside the parent;
  the original of the parent is not inclunded because it should not matter
  */
-NSEdgeInsets BCEdgeInsetsCalculate(NSRect childRect, NSSize parentSize);
```
randomsequence: Do we wait until someone needs this on iOS until we replace it with `BCEdgeInsets`?

Source/BCGeometry.h
samdeane: Yeah, perhaps so. I added them initially because they were being used in logging that I didn't want to have to remove.

If they're going in here though, then you're probably right that they should be guaranteed to do the same thing on both platform, and the inverses should be in there too.

Source/BCGeometry.h
samdeane: Yeah, it's a tricky one to know how much defensive "port it just because it's there" I should do, and how much I should just disable the things that aren't trivial to port, and wait until they're proven to be needed.

I basically decided to mostly do the "just in time" porting route - with the exception of the NSRect/NSPoint/NSSize stuff which was so clearly going to be needed.

Source/BCGeometry.m
samdeane: I have unit tests which prove that currently the two routines do produce the same string. I think it's probably safe to allow them to be the guardian of this assumption.

samdeane: Having said that, right now we don't run the iOS unit tests on Jenkins, so that will eventually need changing.

PR #137, opened 2017-08-02
Code Review: We have BCVoidBlock to replace dispatch_block_t. Let's use it (#14360)

acf:
Code review for We have BCVoidBlock to replace dispatch_block_t. Let's use it (#14360):

Connect to BohemianCoding/Sketch#14360.


Source/BCTypes.h
```diff
@@ -0,0 +1,11 @@
+//  Created by Alan Francis on 15/01/2015.
+//  Copyright (c) 2017 Bohemian Coding. All rights reserved.
+
+#ifndef BCTypes_h
```
samdeane: I think in the world of `#import` rather than `#include` statements, these types of guards are unnecessary.

I don’t know why the Xcode template still puts them in - perhaps to make the C more portable, but I tend to strip them out. Maybe my understanding of `#import` is faulty?

Source/BCTypes.h
```diff
@@ -0,0 +1,11 @@
...
+#define BCTypes_h
+
+// Using block syntax `void (^)()` sucks. The dispatch_block_t is something
+// we've used in the past but its about GCD. So now we have our own
+typedef void(^BCVoidBlock)(void);
```
samdeane: Arguably, this could be just `typedef dispatch_block_t BCVoidBlock`. Not sure if that’s more, or less, clear.

PR #138, opened 2017-08-16
Code Review: Deeply nested menus for overrides from library symbols  (#14301)

opsGavin:
Code review for Deeply nested menus for overrides from library symbols  (#14301):

Connect to BohemianCoding/Sketch#14301.


Source/BCObjectSorting.m
```diff
@@ -0,0 +1,111 @@
...
+
++ (NSArray *)sortObjectsWithCompressedNesting:(NSArray <id<BCSortable>> *)sortableObjects {
+  NSArray *descriptors = [self sortObjectsWithNesting:sortableObjects];
+  return [self consolidateGroups:descriptors.mutableCopy];
+}
+
+
++ (NSString *)normalizedName:(NSString *)name {
+  return [[[name componentsSeparatedByString:@"/"] map:^id(id part) {
+    return [part stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceCharacterSet]];
```
opsGavin: This was originally `[part trimmedString]`, but as that method is currently in Cocolat and it's fairly simple I've just replaced it. Sam's looking at moving some of this stuff to BCFoundation as a separate piece of work, so if I move that method here we're liable to step on one another's toes.

Source/BCObjectSorting.m
```diff
@@ -88,17 +88,23 @@ + (NSMutableArray*)consolidateGroups:(NSMutableArray*)array {
...
       NSString *name = descriptor[BCSortableKeyName];
-      contents = [self consolidateGroups:contents];
-      descriptor = contents[0];
-      name = [NSString stringWithFormat:@"%@/%@", name, descriptor[BCSortableKeyName]];
-      contents = descriptor[BCSortableKeyContents];
-      if (contents) {
-        descriptor = @{ BCSortableKeyName : name,
-                        BCSortableKeyContents : contents };
+      if (contents.count == 1) {
+        contents = [self consolidateGroups:contents];
```
SheffieldKevin: The first line in both the if and else blocks is the same. Is there a reason?

opsGavin: Good spot - there's no reason they shouldn't be a single call outside of the if and that would be neater indeed. Unfortunately I was distracted by something else when you commented on this originally and it's been tested and merged now. I'll keep it on my list if I'm back here again though (which I sure I will be).

